PR #142, opened 2025-01-15
feat: no horizontal scroll

adamrturman:
## Summary

After 45 characters (more than an 0x address and just about the full width of the iframe window) split a long string (like a tx hash) into multiple lines

## Demo 

https://github.com/user-attachments/assets/4572d059-0d50-4463-b484-dfcad9c232f7


src/ChatView.module.css
```diff
@@ -59,6 +59,7 @@
 }
 
 .right {
+  text-align: right;
```
adamrturman, line 62: Current alignment for long messages is to the left

Screenshot 2025-01-16 at 10 59 54 AM

src/sanitize.test.ts
```diff
@@ -116,6 +116,6 @@ This should render anyway
   it("don't render an incomplete markdown link", async () => {
     const input = '[Claude](www.claude.ai';
     const output = await sanitizeContent(input);
-    expect(output).toContain('');
+    expect(output).toBe('');
```
adamrturman, line 119: Stricter assertion

samcode206 (approved):

PR #147, opened 2025-01-16
refactor: consolidate content and role into singular message prop

adamrturman:
## Summary

A message of type `ChatCompletionMessageParam` has two properties `role` and `content`. Rather than passing those as separate props from `Conversation` to `Content` components, let's combine them into a single prop.

### DIFF
--- a/src/Content.tsx
+++ b/src/Content.tsx
@@ -3,18 +3,15 @@ import { sanitizeContent } from './sanitize';
 import styles from './ChatView.module.css';
 import { unmaskAddresses } from './utils/chat/unmaskAddresses';
 import { enforceLineBreak, getStoredMasks } from './utils/chat/helpers';
+import { ChatCompletionMessageParam } from 'openai/resources/index';
 
 export interface ContentProps {
-  content: string;
+  message: ChatCompletionMessageParam;
   onRendered?: () => void;
-  role: string;
 }
 
-export const ContentComponent = ({
-  content,
-  onRendered,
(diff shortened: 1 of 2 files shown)


src/Content.tsx
```diff
@@ -3,18 +3,15 @@ import { sanitizeContent } from './sanitize';
 import styles from './ChatView.module.css';
 import { unmaskAddresses } from './utils/chat/unmaskAddresses';
 import { enforceLineBreak, getStoredMasks } from './utils/chat/helpers';
+import { ChatCompletionMessageParam } from 'openai/resources/index';
 
 export interface ContentProps {
-  content: string;
+  message: ChatCompletionMessageParam;
```
samcode206, line 9: it may be better to keep those props separate like they were before, because message is now an object memo may break because the reference changes but the content and role string may could still be the same, so it leaves a room for less optimal rendering. Basically if you look at conversation history we are looping over a list of messages, when we add a new message, we create a new message array with different message object refs but they hold the same content. if we use an object prop that will cause ContentComponent to rerender every single message, even though the majority of them didn't get any changes in either content, or role.  but becasue the reference to the message object changed it forces ContentComponent to rerender.

samcode206: with content and role being primitive types react can better check if they content of the strings actually changed and need to trigger a rerender.

adamrturman: @samcode206 sounds good

PR #158, opened 2025-01-22
feat: wallet context set up

samcode206:


src/context/WalletContext.tsx
```diff
@@ -0,0 +1,36 @@
...
+  chainId: string;
+  signatureType: SignatureTypes;
+  payload: unknown; // todo: msg? transaction payload? this structure would be different based on the signatureType
+};
+
+export type WalletContextType = {
+  walletAddress: string;
+  walletChainId: string;
+  walletType: WalletTypes;
+  isWalletConnected: boolean;
```
adamrturman, line 28: Nice - now we won't need to do `walletAddress !== ''` in the view layer.

adamrturman (approved):

src/context/WalletContext.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export type SignOpts = {
+  chainId: string;
+  signatureType: SignatureTypes;
+  payload: unknown; // todo: msg? transaction payload? this structure would be different based on the signatureType
+};
+
+export type WalletContextType = {
+  walletAddress: string;
+  walletChainId: string;
```
adamrturman, line 26: Maybe eventually this becomes a white-listed enum of supported chain id's, but this works too.

downing034 (approved):

PR #159, opened 2025-01-22
refactor: remove/comment out unused code to get build passing during refactor

adamrturman:


src/theme/themes.ts
```diff
@@ -1,5 +1,4 @@
 import OrosLogo from '../../src/assets/orosLogo.svg';
-import HardDotFunLogo from '../../src/assets/hardDotFunDiamond.svg';
```
adamrturman, line 2: Removed the image file as well

PR #164, opened 2025-01-23
feat: eth send with updated architecture

adamrturman:


src/services/chain/messages/cosmos/msgSend.ts
```diff
@@ -72,19 +70,7 @@ export class CosmosSendMessage extends CosmosMessageBase<SendToolParams> {
...
-    //     fromAddress,
-    //     toAddress,
-    //     amount: {
-    //       amount,
-    //       denom,
-    //     },
-    //   },
-    // };
-
+    console.log(params);
```
adamrturman, line 73: Satisfy `npm run check` - this will likely get deprecated anyways (unless we want to support sending cosmos assets between kava wallets)

PR #168, opened 2025-01-23
refactor: enums for operation & chain type

adamrturman:

### DIFF
--- a/src/types/chain.ts
+++ b/src/types/chain.ts
@@ -1,6 +1,16 @@
 import { MessageParam } from './messages';
 import { ToolCallStream } from '../toolCallStreamStore';
 
+export enum OperationType {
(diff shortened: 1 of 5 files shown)


src/types/chain.ts
```diff
@@ -26,7 +36,7 @@ export interface ChainOperation {
  */
 export interface ChainMessage extends ChainOperation {
   /** Identifies this as a transaction operation */
-  operationType: 'transaction';
+  operationType: OperationType;
```
adamrturman, line 39: If I fully specify `OperationType.TRANSACTION` here, then in the base implementations, the compiler complains and I'd need to cast

https://github.com/user-attachments/assets/b8556a63-2e7f-46b2-85bf-331a2863aa23

samcode206 (approved):

PR #175, opened 2025-01-23
fix: update token symbols

adamrturman:


src/config/prompts/defaultPrompts.ts
```diff
@@ -38,17 +37,17 @@ You also handle operational tasks tied to tool calls, ensuring all necessary inf
 **User**: "What is my Kava balance?"
 **Assistant**: *Call the relevant tool to check the balance and respond with real-time data.*
 
-**Transactional Query (e.g., msgSend)**:
-**User**: "Send 100 KAVA to address_1"
+**Transactional Query (e.g., EvmTransferMessage)**:
+**User**: "Send 100 USDT to address_1"
```
adamrturman, line 41: Adjusted the example so it will use the correct token symbol for tether.

PR #183, opened 2025-01-24
test: send tx playwright test

adamrturman:
## Summary
Tests pass locally when the funded account is manually set as an already-funded wallet.

https://github.com/user-attachments/assets/d23f737e-efb1-4be8-9ed6-bf94b537c36e

## Todo
Adjust CI/CD to plug in a whale wallet rather than generating and funding multiple accounts


e2e/Metamask.ts
```diff
@@ -0,0 +1,287 @@
...
+  }
+
+  public async switchNetwork() {
+    const metaMaskPage = await this.registerPage.context().newPage();
+    await metaMaskPage.goto(this.extensionLink + 'home.html');
+    await metaMaskPage.waitForSelector('[data-testid="network-display"]', {
+      timeout: 10000,
+    });
+
+    await metaMaskPage.getByTestId('network-display').click();
```
adamrturman, line 181: Everything above this was directly imported from the webapp application.

Everything below is new since we don't have the "Switch Network" button that adjusts to our chain, so we have to manually register the evm chain, then switch to it, then add our existing wallet (which is currently one of the dev wallets with its key in local .env).

e2e/chat.spec.ts
```diff
@@ -1,39 +1,127 @@
...
+describe('chat', () => {
+  test('renders intro message', async ({ page, context }) => {
+    const chat = new Chat(page);
+    await chat.goto();
 
-  expect(chat.messageContainer).not.toBeNull();
+    expect(chat.messageContainer).not.toBeNull();
 
-  await expect(page.getByText('Welcome to Oros!')).toBeVisible();
+    await expect(page.getByText("let's get started").first()).toBeVisible();
```
adamrturman, line 13: `let's get started` appears twice - once as the header and then again in the paragraph that follows

e2e/chat.spec.ts
```diff
@@ -1,39 +1,127 @@
...
+        testInfo.retry,
+      );
+
+      metaMask = await MetaMask.prepareWallet(
+        context,
+        metaMaskExtensionId,
+        walletId,
+        true,
+      );
+    });
```
adamrturman, line 63: To help speed up tests, we only register metamask accounts for tests that will use them.

e2e/chat.spec.ts
```diff
@@ -1,39 +1,127 @@
...
+      await chat.waitForAssistantResponse();
+
+      const messages = await chat.getMessageElementsWithContent();
+      const responseText = await messages[messages.length - 1].innerText();
+
+      //  Verify that the connected wallet has some KAVA balance
+      const kavaMatch = responseText.match(/KAVA: ([\d.]+)/);
+      expect(kavaMatch).toBeTruthy();
+      const kavaBalance = Number(kavaMatch[1]);
+      expect(kavaBalance).toBeGreaterThan(0);
```
adamrturman, line 93: Not sure if we can get a stricter assertion here unless we were to connect a wallet with funds that nobody uses (including other tests)

PR #190, opened 2025-01-27
feat: scroll to bottom of container for inProgress/tx complete cards

samcode206:


src/components/CompleteTxDisplay.tsx
```diff
@@ -1,15 +1,27 @@
...
+  onRendered,
+}: CompleteTxDisplayProps) => {
   const explorerUrl = chain.blockExplorerUrls[0] + 'tx/' + hash;
 
+  useEffect(() => {
+    if (onRendered) {
+      requestAnimationFrame(onRendered);
+    }
+  }, [onRendered]);
+
```
adamrturman, line 24: Maybe we extract this repeated block into a hook?

`useScrollToBottomOnRendered(onRendered)`

adamrturman (approved):
Manually tested the standalone and iframe versions - both look good 🚀

Issue #194, opened 2025-01-28
Prompt hardening: chain balances query


adamrturman:
We are currently *defaulting* to Kava EVM if no chain id is included in the parameters, but I've seen a few unneeded confirmations in the responses from the model

```
User: Check my balances?
System: Can you please specify which chain: Kava EVM, Ethereum, or Kava EVM Internal Testnet
```

Perhaps simply including additional examples in the system prompt will suffice?

PR #199, opened 2025-01-28
feat: add chain name to tx confirmation details

adamrturman:
## Summary
Addresses Issue #194 

- added examples of transacting with other chains
- made the chain name a part of all tx details 

Example issue
User: Send 1 USDT to ...
Assistant: Can you please specify which chain?

## Demo 

https://github.com/user-attachments/assets/1365699e-6bbe-4435-a2cb-d1642a9ca018

### DIFF
--- a/src/config/prompts/defaultPrompts.ts
+++ b/src/config/prompts/defaultPrompts.ts
@@ -37,13 +37,32 @@ You also handle operational tasks tied to tool calls, ensuring all necessary inf
 **User**: "What is my Kava balance?"
 **Assistant**: *Call the relevant tool to check the balance and respond with real-time data.*
 
-**Transactional Query (e.g., EvmTransferMessage)**:
+**General Query**:
+**User**: "What is my USDT balance on Ethereum?"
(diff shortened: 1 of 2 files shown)


src/config/prompts/defaultPrompts.ts
```diff
@@ -37,13 +37,32 @@ You also handle operational tasks tied to tool calls, ensuring all necessary inf
...
+**Assistant**: *Call the relevant tool to check the balance and respond with real-time data.*
+
+**Transactional Message (e.g., EvmTransferMessage)**:
 **User**: "Send 100 USDT to address_1"
 **Assistant**: 
 "I have collected the following information:  
 - Receiving Address: address_1  
 - Token: USD₮  
 - Amount: 100  
+- Chain: Kava EVM
```
adamrturman, line 51: I suppose we don't need to show EVM if its the default chain, but it seems that if crosschain is the goal, it will be coming sooner rather than later.

samcode206 (approved):

PR #201, opened 2025-01-28
fix: validate users balances before transfer

adamrturman:
## Summary
- adds a private method to EvmTransferMessage for validating that a user has enough of a particular denom (on a particular chain) 
- calls this message as part of the validation check

## Before

https://github.com/user-attachments/assets/d9b02f98-ffb7-4af7-9563-cdc3db99a15b

## After

https://github.com/user-attachments/assets/dd19b92e-36b1-4ab7-b646-1a6005e0304d

## Follow ons
- After this, we should refactor the `inProgressComponent` to only render if the `validate` method succeeds. Right now, it shows that the tx is in progress, but then the validation fails and it goes away

### DIFF
--- a/src/services/chain/messages/evm/transfer.ts
+++ b/src/services/chain/messages/evm/transfer.ts
@@ -161,7 +161,6 @@ export class EvmTransferMessage implements ChainMessage<SendToolParams> {
     params: SendToolParams,
     walletStore: WalletStore,
   ): Promise<string> {
-    console.log('calling build transaction');
     const { toAddress, amount, denom } = params;
 
     const { erc20Contracts, rpcUrls, nativeToken } =
--- a/src/services/chain/messages/evm/transfer.ts
+++ b/src/services/chain/messages/evm/transfer.ts
@@ -58,7 +58,10 @@ export class EvmTransferMessage implements ChainMessage<SendToolParams> {
     return InProgressTxDisplay;
   }
 
-  validate(params: SendToolParams, walletStore: WalletStore): boolean {
+  async validate(
+    params: SendToolParams,
+    walletStore: WalletStore,
+  ): Promise<boolean> {
     if (!walletStore.getSnapshot().isWalletConnected) {
       throw new Error('please connect to a compatible wallet');
     }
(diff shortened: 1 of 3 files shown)


src/types/chain.ts
```diff
@@ -33,7 +33,7 @@ export interface ChainOperation<T> {
   walletMustMatchChainID?: boolean;
 
   /** Validates the provided parameters match requirements */
-  validate(params: T, walletStore: WalletStore): boolean;
+  validate(params: T, walletStore: WalletStore): boolean | Promise<boolean>;
```
adamrturman, line 36: We currently have this as a boolean for cosmos side txs, but for eth side we rely on async helpers

src/services/chain/messages/evm/transfer.ts
```diff
@@ -58,7 +58,61 @@ export class EvmTransferMessage implements ChainMessage<SendToolParams> {
...
+      const rawBalance = await contract.balanceOf(address);
+      const formattedBalance = ethers.formatUnits(rawBalance, decimals);
+
+      console.log(
+        formattedBalance,
+        amount,
+        Number(formattedBalance) >= Number(amount),
+      );
+      return Number(formattedBalance) >= Number(amount);
+    } catch (e) {
```
samcode206, line 107: i would either remove the try/catch or rethrow the error in the catch block if we wanna create a custom error instead of the ethers library ones.

src/services/chain/messages/evm/transfer.ts
```diff
@@ -58,7 +58,61 @@ export class EvmTransferMessage implements ChainMessage<SendToolParams> {
...
+        const formattedBalance = ethers.formatUnits(
+          rawBalance,
+          nativeTokenDecimals,
+        );
+        return Number(formattedBalance) >= Number(amount);
+      }
+
+      const erc20Record = getERC20Record(denom, erc20Contracts);
+      if (!erc20Record) {
+        return false;
```
samcode206, line 88: a more descriptive error thrown here would be more useful, otherwise we would fall back to this generic error here https://github.com/Kava-Labs/oros/pull/201/files#diff-e49e9ee75a1f24a197cabb5cc3d100526d6f540e07558642fabc92d93adc88dcR126

samcode206 (approved):

PR #202, opened 2025-01-28
fix: throw errors from validate step

adamrturman:
## Summary
The context-level callback `executeOperation` now has a try-catch block so that we can `setErrorText` with the explicit error. This way, we can provide useful feedback to the user on why their tx couldn't go through.

## Before
UI error doesn't help

https://github.com/user-attachments/assets/af3c3fad-a2e3-43ec-9abd-feb6e7b804dc

And no error is thrown in the console

Screenshot 2025-01-28 at 4 20 38 PM

## After
We should probably do a follow-on to address/remove the "double error" treatment

Perhaps if we receive a response from a failed tool call, 
- don't add it to the message store 
- don't render it to the UI

https://github.com/user-attachments/assets/ace274f6-2dbc-46ff-800e-cd5a0fa5a4bc

### DIFF
--- a/src/context/AppContextProvider.tsx
+++ b/src/context/AppContextProvider.tsx
@@ -72,75 +72,86 @@ export const AppContextProvider = ({
    */
   const executeOperation = useCallback(
     async (operationName: string, params: unknown) => {
-      const operation = registry.get(operationName);
-      if (!operation) {
-        throw new Error(`Unknown operation type: ${operationName}`);
-      }
-
-      let chainId = `0x${Number(2222).toString(16)}`; // default
-      let chainName = ChainNames.KAVA_EVM; // default
-
-      // if operation needs wallet connect
-      // and the current wallet connection isn't one that's included in wantsWallet
-      // we then try to establish that connection
-      if (
-        operation.needsWallet &&
-        Array.isArray(operation.needsWallet) &&
-        !operation.needsWallet.includes(walletStore.getSnapshot().walletType)
-      ) {
-        for (const walletType of operation.needsWallet) {
-          // if chainName exists in params, connect to that chain
-          if (
-            typeof params === 'object' &&
-            params !== null &&
-            chainNameToolCallParam.name in params
-          ) {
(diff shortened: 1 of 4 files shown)


src/context/AppContextProvider.tsx
```diff
@@ -72,75 +72,86 @@ export const AppContextProvider = ({
...
-          params,
-          walletStore,
+        throw new Error('Invalid operation type');
+      } catch (error) {
+        setErrorText(
+          'An error occured: '.concat(
+            error instanceof Error
+              ? error.message
+              : 'An unknown error occurred',
+          ),
```
adamrturman, line 147: Set the error (thrown from within the validate method) in state so we can render it

downing034:
It's kind of funky that in the video the streamed text is above the error that is set. Kind of looks like the thread is out of proper order.

src/context/AppContextProvider.tsx
samcode206: shouldn't the model get the error and show it to the user? i don't think this is ideal, we need to complete the tool call, now if an error occurs the model will not be aware.

samcode206: we just need to figure out why what we had before broke, i would expect the thrown error to be handled in `App.tsx` like it currently is 
```
        const result = await executeOperation(
          name,
          toolCall.function.arguments,
        );
        content = JSON.stringify({
          status: 'ok',
          info: result,
        } as OperationResult);
      } catch (err) {
        content = JSON.stringify({
          status: 'failed',
          info: err,
        } as OperationResult);
```

samcode206: if we catch the error inside executeOperation it will be swallowed and the model will not get a response for it's tool call

adamrturman:
> It's kind of funky that in the video the streamed text is above the error that is set. Kind of looks like the thread is out of proper order.

 I think this will go away when the model returns the error from the tool call instead of just failing - will look into this some more.

adamrturman:
Closing in favor of #204

PR #204, opened 2025-01-28
feat: fix JSON.stringify on instanceof Error

samcode206:


adamrturman (approved):
Manually tested and confirmed that tool call response is the correct error string 👍

PR #205, opened 2025-01-29
feat: Build balances query display

downing034:


src/components/Conversation.tsx
```diff
@@ -91,13 +90,23 @@ const ConversationComponent = ({ messages, onRendered }: ConversationProps) => {
...
+              );
+            } else {
+              return (
+                <CompleteTxDisplay
+                  key={index}
+                  hash={content.info}
+                  chain={chain}
+                />
+              );
+            }
```
adamrturman, line 109: Maybe this conditional could be more direct

```
            if (operation.operationType === OperationType.QUERY) {
              return (
                <CompleteQueryDisplay
                  key={index}
                  content={content.info}
                  onRendered={onRendered}
                />
              );
            } else {
              return (
                <CompleteTxDisplay
                  key={index}
                  hash={content.info}
                  chain={chain}
                />
              );
            }
```

adamrturman (approved):

Issue #203, opened 2025-01-28
bug: in-progress treatment before validation suceeds


adamrturman:
If the user sets up a transaction that will fail the `validate` step, then we see the tx in progress component _a bit too early_

https://github.com/user-attachments/assets/f1c16ad8-4bd6-4a52-bed1-ae79749b86c5

We need the inProgressComponent to be rendered conditionally on success of the validate step.

### Expected Behavior

User: "Send 10 bKAVA to 0x1874c3e9d6e5f7e4f3f22c3e260c8b25ed1433f2"

System: "I have collected the following information:

```
Receiving Address: 0x1874c3e9d6e5f7e4f3f22c3e260c8b25ed1433f2
Token: bKAVA
Amount: 10
Is everything correct?
```

User: "Yes"

System: "Thinking" ➡ "An error occured: failed to find contract address for bKAVA"  // never show the Transaction in Progress Card

adamrturman:
Closed with #206

PR #206, opened 2025-01-29
fix: don't show in progress until validation passes

adamrturman:
## Summary
Addresses Issue #203 

### Happy Path

https://github.com/user-attachments/assets/b14488fd-68cb-4de8-a2c1-9dc1ac69be34

### Error Path

https://github.com/user-attachments/assets/5d6ff118-88f1-4c86-9619-b88f41fe9af4

## Follow On
- It would be nice to not render the empty assistant message that's above the in progress card, causing it to not be aligned with the icon
- For the error path, we see "Thinking" ➡️  "{{empty}}" ➡️ "Thinking" ➡️ "{{error text}}" and it would be nice to eliminate those middle two states

### DIFF
--- a/e2e/chat.spec.ts
+++ b/e2e/chat.spec.ts
@@ -109,11 +109,11 @@ describe('chat', () => {
     //  Confirm the tx
     await chat.submitMessage('Yes');
 
-    //  In progress
-    await expect(page.getByTestId('in-progress-tx-display')).toBeVisible();
-
     const metamaskPopup = await metamaskPopupPromise;
     await confirmMetamaskTransaction(metamaskPopup);
+
+    //  In progress
+    await expect(page.getByTestId('in-progress-tx-display')).toBeVisible();
 
     const provider = new ethers.JsonRpcProvider(
       'https://evm.data.internal.testnet.us-east.production.kava.io',
@@ -153,11 +153,11 @@ describe('chat', () => {
     //  Confirm the tx
     await chat.submitMessage('Yes');
 
-    //  In progress
-    await expect(page.getByTestId('in-progress-tx-display')).toBeVisible();
-
     const metamaskPopup = await metamaskPopupPromise;
     await confirmMetamaskTransaction(metamaskPopup);
+
+    //  In progress
+    await expect(page.getByTestId('in-progress-tx-display')).toBeVisible();
 
     const provider = new ethers.JsonRpcProvider(
       'https://evm.data.internal.testnet.us-east.production.kava.io',
--- a/src/context/AppContextProvider.tsx
+++ b/src/context/AppContextProvider.tsx
@@ -73,11 +73,12 @@ export const AppContextProvider = ({
    */
   const executeOperation = useCallback(
     async (operationName: string, params: unknown) => {
+      setIsOperationValidated(false);
+
       const operation = registry.get(operationName);
       if (!operation) {
         throw new Error(`Unknown operation type: ${operationName}`);
       }
-      setIsOperationValidated(false);
 
       let chainId = `0x${Number(2222).toString(16)}`; // default
       let chainName = ChainNames.KAVA_EVM; // default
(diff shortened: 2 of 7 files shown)


e2e/chat.spec.ts
```diff
@@ -109,12 +109,12 @@ describe('chat', () => {
...
 
-    //  In progress
-    await expect(page.getByTestId('in-progress-tx-display')).toBeVisible();
-
     const metamaskPopup = await metamaskPopupPromise;
     await confirmMetamaskTransaction(metamaskPopup);
 
+    //  In progress
+    await expect(page.getByTestId('in-progress-tx-display')).toBeVisible();
+
```
adamrturman, line 117: This had to change in the automated test because when they click "Yes", we still need to connect their wallet to internal testnet, so they haven't passed the validation step yet, so the in progress component isn't displaying yet.

Makes me think we could possibly repurpose this card to show "Confirm connection in Metamask" if a user is here for the first time and hasn't given permission to connect.

src/context/AppContextProvider.tsx
```diff
@@ -76,6 +77,7 @@ export const AppContextProvider = ({
       if (!operation) {
         throw new Error(`Unknown operation type: ${operationName}`);
       }
+      setIsOperationValidated(false);
```
samcode206, line 80: might be safer to move this up to be the first thing this function does, before an error is potentially thrown. because if an error is thrown before we set the operation to false, it would still show true

src/context/AppContextProvider.tsx
```diff
@@ -46,6 +46,7 @@ export const AppContextProvider = ({
   progressStore: TextStreamStore;
   messageHistoryStore: MessageHistoryStore;
 }) => {
+  const [isOperationValidated, setIsOperationValidated] = useState(false);
```
adamrturman, line 49: I played around with a more OOP-approach of handling whether to render the `inProgressComponent` within the class itself, but it involved much more code than simple react state.

src/useScrollToBottom.tsx
```diff
@@ -1,9 +1,12 @@
...
+export const useScrollToBottom = (
+  onRendered?: () => void,
+  isOperationValidated?: boolean,
+) => {
   useEffect(() => {
     if (onRendered) {
       requestAnimationFrame(onRendered);
     }
-  }, [onRendered]);
+  }, [onRendered, isOperationValidated]);
```
adamrturman, line 11: When operation becomes validated and the component is rendered, be sure to scroll to the bottom so it stays in view.

e2e/chat.spec.ts
adamrturman: https://github.com/user-attachments/assets/e497bd10-bf67-4dde-85f0-bccbca727f34

src/services/chain/messages/evm/transfer.ts
```diff
@@ -55,7 +57,7 @@ export class EvmTransferMessage implements ChainMessage<SendToolParams> {
   ];
 
   inProgressComponent() {
-    return InProgressTxDisplay;
+    return this.hasValidWallet ? InProgressTxDisplay : ConnectWalletPrompt;
```
adamrturman, line 60: Before, we were showing the in progress component immediately, regardless of whether any validation checks had passed. And that was useful to prompt new users to confirm their metamask connection...sort of.

Now, we can determine if the user is making a transfer and they haven't connected a wallet yet, show them a useful prompt to connect metamask.

Then, if that connected wallet passes validation, show the in progress component.


https://github.com/user-attachments/assets/bb9452b8-8346-4ba9-bbd9-2686d5988434

If the wallet doesn't pass the check, never show in progress

https://github.com/user-attachments/assets/4b3c9c17-5f69-4bbb-a027-624b90ce14aa

samcode206 (approved):

downing034 (approved):

PR #208, opened 2025-01-29
feat: suppress follow up after particular tool call

adamrturman:
## Summary
For certain tool calls, we may not want to get a natural language summary of the results, since those results might be nicely formatted UI elements, and we don't want duplicate text summary

If we process a white-listed tool call, use that as a base case to exit the recursive `doChat`.

## Demo 
As a user, I can 
- query balances
- start and cancel a tx
- start and confirm a tx

https://github.com/user-attachments/assets/ee15f6a3-823c-437d-8272-edd21dfe6d8b

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -200,6 +200,7 @@ async function doChat(
   messageStore: TextStreamStore,
   toolCallStreamStore: ToolCallStreamStore,
   executeOperation: ExecuteOperation,
+  shouldSuppressResponse: boolean = false, // Tracks if evm-balances suppressed follow-up
 ) {
   progressStore.setText('Thinking');
   try {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -238,24 +239,28 @@ async function doChat(
...
-        client,
-        messageHistoryStore,
-        tools,
-        progressStore,
-        messageStore,
-        toolCallStreamStore,
-        executeOperation,
-      );
+      // Allow chat to continue only if no evm-balances suppression OR a reset occurred
+      if (!nextShouldSuppressResponse) {
```
samcode206, line 251: ideal when the model requests a tool call, we do give it the response, not sure how this would affect subsequent tool calls or if it's okay to not have a response when a model requests a tool call.

adamrturman: Open AI's schema is to summarize the results of a tool call with natural language, so we are kinda implementing an anti-pattern, but I think it makes sense for this use case.

We are still getting the result of the tool call, like we would for cancelling a tx in metamask.


Screenshot 2025-01-29 at 3 48 39 PM

adamrturman: It's kinda like we are adding another base case to exit the recursion.

adamrturman:
Closing in favor of #209

PR #220, opened 2025-02-03
refactor: move tool registry onto model config object

adamrturman:


src/services/modelRegistry.ts
```diff
@@ -0,0 +1,22 @@
+import { initializeRegistry } from '../context/initializeRegistry';
+
+const registry = initializeRegistry();
+
+export const MODEL_REGISTRY = {
+  blockchain: {
+    'gpt-4o': {
+      name: 'gpt-4o',
+      tools: registry.getToolDefinitions(),
+    },
```
downing034, line 10: Let's add a: `description` key to this as well. The mocks have a "short description" in the selector.  I think we should also add/build a `systemPrompt` key and set it up `getSystemPrompt()` for the models that need them

samcode206 (approved):

src/services/modelRegistry.ts
```diff
@@ -0,0 +1,96 @@
...
+    };
+  };
+}
+
+interface ModelRegistry {
+  blockchain: Record<SupportedBlockchainModels, ModelConfig>;
+  reasoning: Record<SupportedReasoningModels, ModelConfig>;
+}
+
+const registry = initializeRegistry();
```
downing034, line 45: let's rename this or rename `initializeRegistry` to be more specific. Maybe `const messageRegistry = initializeMessageRegistry();`

downing034 (approved):

PR #222, opened 2025-02-04
feat: rebuild model config and app context config

downing034:


src/features/blockchain/config/models.ts
```diff
@@ -1,39 +1,29 @@
...
-  walletStore: WalletStore,
-) => {
-  /**
-   * Executes a chain operation with the provided parameters.
-   * Handles both transaction and query operations.
-   * @param operationType - Type identifier for the operation
-   * @param params - Parameters for the operation
-   * @returns Result of the operation (transaction or query result)
-   */
+export const messageRegistry = initializeMessageRegistry();
```
downing034, line 21: I think this may be causing a bug and could use a second set of eyes on it. This initializes fine, but I don't think it gets updated like it used to when it was in AppContextProvider, so now when you type: `send 1 USDT to address_1`, the InProgress component doesn't render as expected.

If this is indeed the issue, then I think there may be a few options:

1. Move it into some kind of hook that can be called
2. Maybe we can add a subscription to the walletStore somewhere:
```
const walletSnapshot = useSyncExternalStore(
  walletStore.subscribe,
  walletStore.getSnapshot,
);
```
3) Find a way to update / force an update to `createBlockchainOperations` so it rebuilds (a hook would sort of do this too).

If it really boils down to it, I suppose it could be moved back into the AppContextProvider, but I'd like to avoid that to keep the proper separation of concerns between models and core code

src/core/components/ToolCallProgressCards.tsx
```diff
@@ -12,12 +13,11 @@ export const ToolCallProgressCards = ({
     toolCallStreamStore.subscribe,
     toolCallStreamStore.getSnapShot,
   );
-  const { registry } = useAppContext();
 
   if (!toolCallStreams.length) return null;
 
   return toolCallStreams.map((toolCall) => {
-    const operation = registry.get(toolCall.function.name ?? '');
+    const operation = messageRegistry.get(toolCall.function.name ?? '');
```
samcode206, line 20: @downing034 is it this component that is not showing?

adamrturman: @samcode206 - it's the in progress tx display that doesn't show when send

src/core/context/AppContextProvider.tsx
```diff
@@ -0,0 +1,65 @@
...
+    getModelConfig('gpt-4o-mini'),
+  );
+
+  // This callback would be passed to components that need to switch models
+  const handleModelChange = useCallback((modelName: string) => {
+    const newConfig = getModelConfig(modelName);
+    setModelConfig(newConfig);
+  }, []);
+
+  const operations =
```
samcode206, line 40: this will call createOperations on every rerender of the App context. not sure we would want that

src/core/context/AppContextProvider.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <AppContext.Provider
+      value={{
+        messageHistoryStore,
+        messageStore,
+        progressStore,
+        walletStore,
+        toolCallStreamStore,
+        modelConfig,
+        handleModelChange,
+        ...operations, // Spread operations if they exist
```
samcode206, line 53: how deeply nested is this operation object? spread may not always work for deeply nested objects

samcode206: also why do we need to spread this, why cant we pass the entire operation value.

src/core/components/ToolCallProgressCards.tsx
samcode206: i am not seeing this class having the inProgressComponent set up anymore when i log it 
Screen Shot 2025-02-04 at 11 06 28 AM

src/features/blockchain/config/models.ts
```diff
@@ -124,11 +110,53 @@ export const useExecuteOperation = (
...
+export const blockchainModels: Record<string, BlockchainModelConfig> = {
+  'gpt-4o': {
+    name: 'gpt-4o',
+    description:
+      'A robust and powerful model for executing blockchain-specific actions',
+    tools: messageRegistry.getToolDefinitions(),
+    systemPrompt: messageRegistry.getSystemPrompt(),
+    introText: messageRegistry.getIntroText(),
+    components: {
+      transaction: {
```
samcode206, line 133: what those components here and how do they differ from the ones defined in the `EvmTransferMessage`'s inProgressComponent

downing034:
This PR has been broken into smaller PRs that were merged. The last remaining items had a bug in them and can be moved at a separate time as they are not pertinent to the application working

PR #225, opened 2025-02-04
feat: conditionally rendered navbar

adamrturman:
## Standalone (feature enabled)

https://github.com/user-attachments/assets/cbe44316-a41f-484b-8927-428533fef89e

## Standalone (feature disabled)
Screenshot 2025-02-04 at 10 55 07 AM

## Embedded iframe (pointed at localhost)

Screenshot 2025-02-04 at 10 56 23 AM


src/components/ChatView.tsx
```diff
@@ -8,6 +8,8 @@ import { maskAddresses } from '../utils/chat/maskAddresses';
 import { getStoredMasks, updateStoredMasks } from '../utils/chat/helpers';
 import { useAppContext } from '../context/useAppContext';
 
+const FEAT_UPDATED_DESIGN = import.meta.env.VITE_FEAT_UPDATED_DESIGN;
+
```
adamrturman, line 12: Which deployed environment do we want to set this to `true`?

src/components/ChatView.tsx
```diff
@@ -102,8 +104,15 @@ export const ChatView = ({
...
   }, []);
 
+  const isInIframe = window !== window.parent;
+
+  const showNavBar = !isInIframe && FEAT_UPDATED_DESIGN;
+
   return (
-    <div id={styles.chatview} data-testid="chatview">
+    <div
+      id={showNavBar ? styles.updatedChatView : styles.chatview}
```
adamrturman, line 113: For the updated version, we don't need the chat to be 100% view height because that will push the header out of view.

src/components/NavBar.module.css
```diff
@@ -0,0 +1,58 @@
+.container {
+  display: flex;
+  flex-direction: column;
+  background-color: #111111;
```
adamrturman, line 4: I'll do a follow-on PR to use the theme colors, but I don't want to block other work.

samcode206 (approved):

src/components/NavBar.tsx
```diff
@@ -0,0 +1,43 @@
+import styles from './NavBar.module.css';
```
downing034, line 1: This component (and css file) are in the wrong directory. They should be in `src/core/components/`

src/features/blockchain/services/messages/evm/transfer.ts
```diff
@@ -112,6 +112,7 @@ export class EvmTransferMessage implements ChainMessage<SendToolParams> {
     params: SendToolParams,
     walletStore: WalletStore,
   ): Promise<boolean> {
+    console.log('hasvalidwallet', this.hasValidWallet);
```
downing034, line 115: Is this needed still?

src/shared/assets/DeepseekIcon.tsx
```diff
@@ -0,0 +1,25 @@
+const DeepseekIcon = () => {
```
downing034, line 1: wondering if this should be moved to `features/reasoning/assets` instead.

PR #228, opened 2025-02-04
feat: move out reasoning config to it's proper feature dir

downing034:

### DIFF
--- a/src/features/reasoning/config/prompts/defaultPrompts.ts
+++ b/src/features/reasoning/config/prompts/defaultPrompts.ts
@@ -1,0 +1,7 @@
+export const defaultIntroText = 'What can I help you with?';
+
+export const defaultSystemPrompt =
+  'You are an AI assistant designed to provide helpful, accurate, and relevant responses to user queries across various topics. Be concise when needed and detailed when appropriate.';
+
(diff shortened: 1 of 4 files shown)


src/features/reasoning/config/prompts/defaultPrompts.ts
```diff
@@ -0,0 +1,7 @@
+export const defaultIntroText = 'What can I help you with?';
+
+export const defaultSystemPrompt =
+  'You are an AI assistant designed to provide helpful, accurate, and relevant responses to user queries across various topics. Be concise when needed and detailed when appropriate.';
+
+export const defaultCautionText =
+  'KavaAI can make mistakes. Check important info.';
```
adamrturman, line 7: Thought about this before - if this doesn't change between models/agents, then I think it can come off the config and just be part of the input component on the view level?

adamrturman (approved):

src/features/reasoning/config/prompts/defaultPrompts.ts
downing034: agree. It is unclear at the moment if that will be the case, so going to keep it like this to be the least invasive and similar to the blockchain config.

PR #231, opened 2025-02-04
feat: model switcher dropdown UI

adamrturman:
## Summary
Note this just gets the UI on the page. The following two issues will be addressed in a follow on.

- Disabling the model switcher when a chat is in progress will be handled in an immediate follow-on.

- Disable switching to blockchain model on mobile.]

Then we will need to wire this into set model config in context. I propose that we simply keep track of the string model name `chat-gpt-4o` and then use that string to look up the rest of the config from the `MODEL_REGISTRY`. 

That will make it simpler in the view layer so when we select the name  of the chosen model (to show as primary in the UI) we are also setting the backend state as well.

## Demo

https://github.com/user-attachments/assets/60a78016-6586-4f6a-92d7-b62340c16fdb


src/core/components/NavBar.tsx
```diff
@@ -6,12 +7,30 @@ import { isInIframe } from '../utils/isInIframe';
...
   const showNavBar = !isIframe && FEAT_UPDATED_DESIGN;
+  const [dropdownOpen, setDropdownOpen] = useState(false);
+  const [selectedModel, setSelectedModel] = useState(modelOptions[0]);
 
   if (!showNavBar) return null;
 
+  const handleSelect = (model: ModelOption) => {
+    setSelectedModel(model);
+    setDropdownOpen(false);
+  };
```
adamrturman, line 32: Eventually, we will also be able to set the model name which we can use to key into the `MODEL_REGISTRY` to get system prompt, tools, components, etc.

src/core/components/NavBar.tsx
```diff
@@ -6,12 +7,30 @@ import { isInIframe } from '../utils/isInIframe';
...
+interface ModelOption {
+  id: string;
+  name: string;
+  icon: JSX.Element;
+}
+
+const modelOptions: ModelOption[] = [
+  { id: 'deepseek', name: 'DeepSeek RI 67TB', icon: <DeepseekIcon /> },
+  { id: 'kavaai', name: 'Kava AI', icon: <>Logo coming soon</> },
+];
```
adamrturman, line 19: Just placeholder for now.

PR #232, opened 2025-02-04
feat: mobile chat history sidebar

samcode206:
## Demo 

https://github.com/user-attachments/assets/4de8bdc8-fb3a-45b0-9010-65c8cbba77fb


downing034:
The media query stuff here is nice. This PR however is different from figma when the mobile menu appears. The start new chat icon changes on mobile to a paper/pencil icon (like what's on gpt). instead of the large rectangular button

src/App.tsx
```diff
@@ -153,22 +160,34 @@ export const App = () => {
     ]);
   }, [handleCancel, messageHistoryStore, modelConfig.systemPrompt]);
 
+  const isMobile = useMediaQuery('(max-width: 768px)');
```
adamrturman, line 163: Maybe this could become a hook like useIsMobile so it's easy to remember, import, and keep consistent across the app?

src/shared/theme/useMediaQuery.tsx
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useState } from 'react';
+
+export const useMediaQuery = (query: string) => {
```
adamrturman, line 3: Nice 😎

src/App.tsx
samcode206: thanks i will create a wrapper hook `useIsMobile` that uses this hook underneath

samcode206:
notes: 
* we need to swap the hamburger icon for a `x` icon when it's toggled and the navbar is open on mobile
* on mobile we need to conditionally not render the new chat button and instead render the icon version of it on the right side of the navbar

src/shared/theme/useMediaQuery.tsx
```diff
@@ -0,0 +1,21 @@
...
+    };
+
+    matchQuery.addEventListener('change', onChange);
+    return () => matchQuery.removeEventListener('change', onChange);
+  }, [query]);
+
+  return matches;
+};
+
+
```
downing034, line 21: should fix this warning as well. Looks like just an extra line at the end of the file

downing034 (approved):

src/shared/theme/useMediaQuery.tsx
adamrturman: > should fix this warning as well. Looks like just an extra line at the end of the file

It might be useful to have things like this cleaned up on commit via husky.

src/App.tsx
```diff
@@ -154,22 +161,41 @@ export const App = () => {
     ]);
   }, [handleCancel, messageHistoryStore, modelConfig.systemPrompt]);
 
+  const isMobile = useIsMobile();
+  const [burgerToggle, setBurgerToggle] = useState(false);
```
adamrturman, line 165: I think if this is named something like `[chatHistoryOpen, setChatHistory]`, it makes its function a bit more clear.

adamrturman (approved):

PR #234, opened 2025-02-04
feat: handle disabled states for dropdown

adamrturman:
## Summary
- if a user clicks outside of the dropdown it closes
- a user can use the arrow keys to toggle options
- the user can use the enter and escape keys to manipulate the dropdown
- the dropdown is disabled if a user has started a chat, but becomes enabled again when history is cleared

## Demo 

https://github.com/user-attachments/assets/526a6c31-5b19-48ac-ad3a-94404f6d55da

## Todo
- [x] testing
- [x] componentization/refactor
- [ ] move key handling/outside click to hooks

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -86,6 +86,7 @@ jobs:
       VITE_E2E_WALLET_KEY: ${{ secrets.VITE_E2E_WALLET_KEY }}
       VITE_E2E_WALLET_MNEMONIC: ${{ secrets.VITE_E2E_WALLET_MNEMONIC }}
       VITE_E2E_WALLET_ADDRESS: ${{ secrets.VITE_E2E_WALLET_ADDRESS }}
+      VITE_FEAT_UPDATED_DESIGN: true
     outputs:
       changed: ${{ steps.diff.outputs.diff != '' }}
       image-name: ${{ steps.image-name.outputs.image }}
--- a/e2e/chat.spec.ts
+++ b/e2e/chat.spec.ts
@@ -2,6 +2,7 @@ import { describe, expect, test, retryButtonClick } from './fixtures';
 import { Chat } from './Chat';
 import { MetaMask } from './Metamask';
 import { ethers } from 'ethers';
+import { devices } from '@playwright/test';
 
 describe('chat', () => {
   test('renders intro message', async ({ page }) => {
@@ -202,5 +203,30 @@ describe('chat', () => {
     await chat.submitMessage('test message');
 
     await expect(modelButton).toBeDisabled();
+  });
+  test('model dropdown interactions in mobile', async ({ browser }) => {
+    const context = await browser.newContext({
+      ...devices['iPhone 13'],
+    });
+
+    const page = await context.newPage();
+
+    const chat = new Chat(page);
+    await chat.goto();
+
+    const modelButton = page.getByRole('button', { name: 'Select model' });
+    await modelButton.click();
+
+    const gpt4o = page.getByRole('option').filter({ hasText: /^gpt-4o$/ });
+    const gpt4oMini = page
+      .getByRole('option')
+      .filter({ hasText: /^gpt-4o-mini$/ });
+    const deepseek = page.getByRole('option').filter({ hasText: 'deepseek' });
+
+    await expect(gpt4o).toBeDisabled();
+    await expect(gpt4oMini).toBeDisabled();
+    await expect(deepseek).toBeEnabled();
+
+    await context.close();
   });
 });
--- a/e2e/chat.spec.ts
+++ b/e2e/chat.spec.ts
@@ -3,6 +3,12 @@ import { Chat } from './Chat';
 import { MetaMask } from './Metamask';
 import { ethers } from 'ethers';
 import { devices } from '@playwright/test';
+import {
+  DEFAULT_MODEL_NAME,
+  NUMBER_OF_SUPPORTED_MODELS,
+} from '../src/core/config/models';
+import { blockchainModels } from '../src/features/blockchain/config/models';
+import { reasoningModels } from '../src/features/reasoning/config/models';
 
 describe('chat', () => {
   test('renders intro message', async ({ page }) => {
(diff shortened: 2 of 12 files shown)


e2e/chat.spec.ts
```diff
@@ -164,4 +164,43 @@ describe('chat', () => {
 
     expect(formattedAmount).toBe(0.2345);
   });
+  test('model dropdown interactions', async ({ page }) => {
+    const NUMBER_OF_SUPPORTED_MODELS = 3;
+    const DEFAULT_MODEL_NAME = 'gpt-4o-mini';
```
adamrturman, line 169: These will likely change, but at least there's only one place to update in these tests when that happens.

.github/workflows/ci-cd.yml
```diff
@@ -86,6 +86,7 @@ jobs:
       VITE_E2E_WALLET_KEY: ${{ secrets.VITE_E2E_WALLET_KEY }}
       VITE_E2E_WALLET_MNEMONIC: ${{ secrets.VITE_E2E_WALLET_MNEMONIC }}
       VITE_E2E_WALLET_ADDRESS: ${{ secrets.VITE_E2E_WALLET_ADDRESS }}
+      VITE_FEAT_UPDATED_DESIGN: true
```
adamrturman, line 89: Added so we can test these new features in playwright in ci cd

src/core/components/ModelDropdownExpanded.tsx
```diff
@@ -0,0 +1,38 @@
...
+  setDropdownOpen: (dropdownOpen: boolean) => void;
+}
+
+const ModelDropdownExpanded = ({ setDropdownOpen }: ModelDropdownProps) => {
+  const { handleModelChange } = useAppContext();
+
+  const handleModelClick = (model: ModelConfig) => {
+    handleModelChange(model.id);
+    setDropdownOpen(false);
+  };
```
adamrturman, line 16: @downing034 - one change I made here is closing the dropdown when the user selects a model.

src/core/components/ModelDropdownButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const hasUserMessages = messages.length > 1;
+
+  useEffect(() => {
+    if (hasUserMessages) {
+      setIsDisabled(true);
+      setDropdownOpen(false);
+    } else {
+      setIsDisabled(false);
+    }
+  }, [hasUserMessages, setDropdownOpen]);
```
adamrturman, line 37: I don't think the user can actually enter a chat with the dropdown open, because as soon as they click outside the dropdown, it will close. But just to be safe let's be sure its closed.

src/core/components/ModelDropdownExpanded.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <div className={styles.dropdownMenu} data-testid="model-dropdown-menu">
```
adamrturman, line 18: Added a test id to verify open vs. closed in playwright

src/core/components/ModelDropdownExpanded.tsx
```diff
@@ -0,0 +1,66 @@
...
+    };
+
+    matchQuery.addEventListener('change', onChange);
+    return () => matchQuery.removeEventListener('change', onChange);
+  }, [query]);
+
+  return matches;
+};
+
+export const useIsMobile = () => useMediaQuery('(max-width: 768px)');
```
adamrturman, line 30: Will remove once #232 is in

src/core/components/ModelDropdownExpanded.tsx
```diff
@@ -0,0 +1,66 @@
...
+    handleModelChange(model.id);
+    setDropdownOpen(false);
+  };
+  return (
+    <div className={styles.dropdownMenu} data-testid="model-dropdown-menu">
+      {getAllModels().map((model) => {
+        const ModelOptionIcon = model.icon;
+
+        //  since we don't support metamask on mobile, don't allow a user to switch to blockchain models
+        const disableOption = isMobile && isBlockchainModelName(model.id);
```
adamrturman, line 47: Keep users from choosing a blockchain model if they are on mobile until we support a mobile wallet.

e2e/chat.spec.ts
downing034: Maybe we can move `DEFAULT_MODEL_NAME` into a config in the core directory. The AppContextProvider also sort of sets a "default" selected model and that is where I'd expect it to change.

If we change the default behaviour there, I would expect the e2e test to have to be updated as well. If we set this here we run the risk of having the test pass when the default changes in app context which has the potential to hide bugs.

src/core/components/NavBar.module.css
```diff
@@ -90,6 +90,15 @@
...
 }
 
+.dropdownItem:disabled {
+  opacity: 0.5;
+  cursor: not-allowed;
+  background-color: #1a1a1a;
+}
+
+.dropdownItem:disabled:hover {
+  background-color: #1a1a1a;
```
downing034, line 100: I noticed this earlier, but this PR includes custom css vs using the theme styling. I think we may need to plug this into the theme instead, though I am not sure if the current theme is good enough to just modify this.

We're starting this with dark mode, but there was talk of a light mode as well at some point.

src/App.tsx
```diff
@@ -175,10 +175,7 @@ export const App = () => {
     <>
       {isReady && (
         <div className={styles.appContent}>
-          <NavBar
-            chatHistoryOpen={chatHistoryOpen}
```
adamrturman, line 179: chatHistoryOpen prop was not being used in the NavBar

src/core/components/NavBar.module.css
adamrturman: I'll update this file to use stuff from the theme. Well, Claude will 😎

e2e/chat.spec.ts
```diff
@@ -164,4 +165,66 @@ describe('chat', () => {
 
     expect(formattedAmount).toBe(0.2345);
   });
+  test('model dropdown interactions', async ({ page }) => {
+    const DEFAULT_MODEL_NAME = 'gpt-4o-mini';
+    const NUMBER_OF_SUPPORTED_MODELS = 3;
```
adamrturman, line 170: Importing these values from `core/components/config` causes a weird css-module related bug: https://github.com/Kava-Labs/oros/actions/runs/13165276631/job/36743766762

downing034 (approved):

PR #238, opened 2025-02-05
poc: use LLM to generate summary of chat

adamrturman:
## Summary

Not intended to be merged, but can be used as a basis for implementation when we want to go beyond just simply putting the first however many characters of the user's first message. 

## Demo

User: "Can you help me send KAVA to another address?"
Summary: ""Assistance with Kava Transactions and Balance Inquiries"

### DIFF
--- a/src/core/utils/useSummarizeChat.ts
+++ b/src/core/utils/useSummarizeChat.ts
@@ -1,0 +1,44 @@
+import { useState, useEffect } from 'react';
+import OpenAI from 'openai';
+import { ChatCompletionMessageParam } from 'openai/resources/index';
+
+export const useSummarizeChat = (
+  client: OpenAI,
+  messages: ChatCompletionMessageParam[],
+): string => {
+  const [summary, setSummary] = useState('');
+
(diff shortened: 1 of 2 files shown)


src/core/utils/useSummarizeChat.ts
```diff
@@ -0,0 +1,44 @@
...
+        console.error('Error generating summary:', error);
+        const firstMessage = initialMessages.find(
+          (m) => m.role === 'user',
+        )?.content;
+        setSummary(firstMessage?.slice(0, 57) + '...' || 'New Chat');
+      }
+    };
+
+    generateSummary();
+  }, [messages, client]);
```
samcode206, line 41: one problem with this is that the messages array's reference changes completely everytime we add a new message entry to it. so as soon as the messages array has some length greater than 1 it will call fetch and try to summarize everytime we add a new message.

adamrturman: Sure - feel free to adjust this code or start from scratch. Just meant as a template/POC

PR #243, opened 2025-02-05
feat: deepseek is the default model

adamrturman:
## Summary
As per mocks, deepseek is the default model. Also, its the only one supported on mobile.

Screenshot 2025-02-05 at 3 27 38 PM

### DIFF
--- a/e2e/chat.spec.ts
+++ b/e2e/chat.spec.ts
@@ -166,13 +166,13 @@ describe('chat', () => {
     expect(formattedAmount).toBe(0.2345);
   });
   test('model dropdown interactions', async ({ page }) => {
-    const DEFAULT_MODEL_NAME = 'gpt-4o-mini';
+    const DEFAULT_MODEL_DISPLAY_NAME = 'DeepSeek RI 67TB';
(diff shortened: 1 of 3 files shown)


e2e/chat.spec.ts
```diff
@@ -9,6 +9,9 @@ describe('chat', () => {
     const chat = new Chat(page);
     await chat.goto();
 
+    //  todo - remove when the default (reasoning) model is functioning
+    await chat.switchToBlockchainModel();
+
```
adamrturman, line 14: We need to switch to the blockchain model (for now) for all e2e tests, because deepseek isn't wired in yet.

Once that happens, we can remove these from the simple text reply tests.

e2e/chat.spec.ts
```diff
@@ -86,6 +94,8 @@ describe('chat', () => {
     const chat = new Chat(page);
     await chat.goto();
 
+    await chat.switchToBlockchainModel();
+
```
adamrturman, line 98: However, we will keep it around for tests that do blockchain queries and transactions.

downing034 (approved):

PR #249, opened 2025-02-06
feat: update input placement

adamrturman:
## Summary
- updates input to move from centered location before chat starts to bottom of page when chat begins
- breaks out chat view into smaller components

## Demo

https://github.com/user-attachments/assets/44b3fcc9-8877-48b7-82b5-9e5227c97ebe


adamrturman:
Closing in favor of #250

PR #251, opened 2025-02-06
style: fill available

adamrturman:
## Before
Screenshot 2025-02-06 at 12 43 04 PM

## After 
Screenshot 2025-02-06 at 12 42 35 PM

### DIFF
--- a/src/core/components/ChatView.module.css
+++ b/src/core/components/ChatView.module.css
@@ -280,7 +280,9 @@
 }
 
 .inputRaised {
-  width: -webkit-fill-available;
+  width: -webkit-fill-available; /* WebKit browsers (Chrome, Edge, Safari) */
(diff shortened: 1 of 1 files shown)


samcode206 (approved):

src/core/components/ChatView.module.css
```diff
@@ -280,6 +280,9 @@
 }
 
 .inputRaised {
+  width: -webkit-fill-available; /* WebKit browsers (Chrome, Edge, Safari) */
+  width: -moz-available;
+  width: fill-available;
```
downing034, line 285: This is interesting. Previously did we used to have a fixed width on this? Curious did this change because of the new chat view? I'd almost say keep the input a consistent width the entire way instead of supporting the change

PR #253, opened 2025-02-06
feat: Reconfigure new chat location

downing034:
- Add New Chat to Nav bar on mobile
- Hide New Chat button in history sidebar on mobile


src/core/components/ChatHistory.tsx
```diff
@@ -37,25 +40,16 @@ export const ChatHistory = () => {
 
   return (
     <div className={styles.chatHistoryContainer}>
-      <button
-        onClick={() => messageHistoryStore.reset()}
-        className={styles.newChatButton}
-        data-testid="new-chat-container-button"
-      >
-        <svg
```
downing034, line 45: This diff formatted funny. All I did here was take out the <svg> code and replaced it with the exact same SVG code as a component: NewChatIcon since that will need to be used in multiple places now

samcode206 (approved):

src/core/components/NavBar.module.css
```diff
@@ -28,7 +28,7 @@
 .logo {
   display: flex;
   align-items: center;
-  padding: var(--spacing-md) var(--spacing-xl) var(--spacing-md) 0;
+  padding: var(--spacing-md) 0 var(--spacing-md) 0;
```
downing034, line 31: The code here previously was a temporary adjustment to fill space until this NewChatIcon was added. Now it's adjusted to have correct spacing on the edges and space correctly with the other elements

PR #275, opened 2025-02-07
feat: if in iframe default model to blockchain

downing034:


samcode206 (approved):

src/core/context/AppContextProvider.tsx
```diff
@@ -38,8 +38,10 @@ export const AppContextProvider = ({
     initializeMessageRegistry(),
   );
 
+  const isInIframe = window !== window.parent;
```
adamrturman, line 41: You can call the function isInIframe() instead.

adamrturman (approved):

PR #278, opened 2025-02-08
feat: show reset chat in prod and apply go-live

downing034:

### DIFF
--- a/src/core/components/ChatView.tsx
+++ b/src/core/components/ChatView.tsx
@@ -117,10 +117,7 @@ export const ChatView = ({
       >
         {hasMessages && (
           <>
-            {/* Important: When this goes live, the conditional to render ResetChatIcon will change.
-                       We only want to show this trash can in the iframe: !FEAT_UPDATED_DESIGN wil change to  isIframe
-             */}
-            {!FEAT_UPDATED_DESIGN && (
+            {isIframe && (
               <div id={styles.stickyHeader}>
                 <button
                   id={styles.resetButton}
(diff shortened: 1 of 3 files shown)


src/core/components/Conversation.tsx
```diff
@@ -32,17 +33,11 @@ const ConversationComponent = ({ messages, onRendered }: ConversationProps) => {
...
-       - update the #conversation id in chat view to have:  var(--spacing-lg);
-       - remove style={ on the div}}
-    
-    */
   return (
     <div
       id={styles.conversation}
       data-testid="conversation"
-      style={{ marginTop: FEAT_UPDATED_DESIGN ? '24px' : '-10px' }}
+      style={{ marginTop: !isInIframe() ? '24px' : '-10px' }}
```
downing034, line 40: my github comment above was not correct. We still need the negative margin in the iframe. Remove the css from the modules for marginTop and just apply it here as a style for now

src/core/components/ChatView.tsx
```diff
@@ -117,10 +117,7 @@ export const ChatView = ({
       >
         {hasMessages && (
           <>
-            {/* Important: When this goes live, the conditional to render ResetChatIcon will change.
-                       We only want to show this trash can in the iframe: !FEAT_UPDATED_DESIGN wil change to  isIframe
-             */}
-            {!FEAT_UPDATED_DESIGN && (
+            {isIframe && (
```
downing034, line 120: need to get the trash bin in now that we're live

PR #280, opened 2025-02-10
test: chat history e2e coverage

adamrturman:
## Summary
- test coverage for switching between conversations and verifying that local storage populates


e2e/Chat.ts
```diff
@@ -79,22 +78,6 @@ export class Chat {
...
-          return true;
-        }
-        return false;
-      },
-      {
-        timeout: 40000,
-      },
-    );
-  }
-
```
adamrturman, line 97: Removed unused methods

PR #284, opened 2025-02-10
feat: conversation history search function

adamrturman:
## Summary
- creates a unit-tested, currently unimplemented helper that takes in the the object from local storage conversation history 
and returns an array of the matching conversation's id & title (might only need id?)

### DIFF
--- a/src/core/utils/searchConversationHistory.ts
+++ b/src/core/utils/searchConversationHistory.ts
@@ -1,0 +1,33 @@
+import { ConversationHistory } from '../context/types';
+
+interface SearchResult {
+  id: string;
+  title: string;
+}
+
+/**
+ * Searches through conversation history for matches in titles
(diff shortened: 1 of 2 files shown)


src/core/utils/searchConversationHistory.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Searches through conversation history for matches in titles
+ * @param searchTerm - The term to search for
+ * @param conversations - The conversation store object
+ * @returns Array of matching conversation IDs and titles
+ */
+export function searchConversationHistory(
+  searchTerm: string,
+  conversations: Record<string, ConversationHistory>,
+): SearchResult[] {
```
adamrturman, line 17: I can envision this function being run on input change (maybe after input.length > 3 so that we don't get immediate matches for the first letter?) as a user types to search for conversation title

PR #296, opened 2025-02-10
style: no horizontal scroll on Safari

adamrturman:

### DIFF
--- a/src/App.module.css
+++ b/src/App.module.css
@@ -15,3 +15,9 @@
     display: block;
   }
 }
+
+html, body {
+  overflow-x: hidden;
+  width: 100%;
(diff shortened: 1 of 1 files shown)


samcode206 (approved):

src/App.module.css
```diff
@@ -15,3 +15,9 @@
     display: block;
   }
 }
+
+html, body {
+  overflow-x: hidden;
+  width: 100%;
+  max-width: 100vw;
```
downing034, line 22: I swear we had this set somewhere else to prevent it which seemed to be working on FF.

downing034 (approved):

PR #300, opened 2025-02-10
style: fix new chat button alignment between text and icon

adamrturman:
## Summary

Screenshot 2025-02-10 at 4 13 01 PM

### DIFF
--- a/src/core/components/ChatHistory.module.css
+++ b/src/core/components/ChatHistory.module.css
@@ -61,3 +61,8 @@
     background-color: var(--colors-bgTertiary);
   }
 }
+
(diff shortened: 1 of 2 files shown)


downing034:
I know the code has it displayed center, but it looks like it is slightly aligned to the top. Is there a margin or padding on the image that would off-vertically-center the text to be slightly higher?

adamrturman:
> I know the code has it displayed center, but it looks like it is slightly aligned to the top. Is there a margin or padding on the image that would off-vertically-center the text to be slightly higher?

I made a minor tweak that improves aligning the text and the svg view box.

PR #304, opened 2025-02-10
feat: add tooltip for disabled states in model selection

adamrturman:
## Summary
- creates a reusable, positionable tooltip component for handling the disabled state of the model selector dropdown (when a conversation has started) and for switching to a blockchain model on mobile

## Demo 

https://github.com/user-attachments/assets/1aafe8c7-7ae0-4321-98e0-793b0ef82acf

### DIFF
--- a/src/core/components/Tooltip.tsx
+++ b/src/core/components/Tooltip.tsx
@@ -1,0 +1,44 @@
+import React, { useState, useRef, useEffect } from 'react';
+import styles from './Tooltip.module.css';
+
+interface TooltipProps {
+  content: string;
+  children: React.ReactNode;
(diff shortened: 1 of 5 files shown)


src/core/components/Tooltip.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { useState, useRef, useEffect } from 'react';
+import styles from './Tooltip.module.css';
+
+interface TooltipProps {
+  content: string;
+  children: React.ReactNode;
+  topMargin: string;
```
adamrturman, line 7: Since we are building this from scratch, it will be useful to be able to manually position the tooltip.

PR #306, opened 2025-02-11
feat: rendering thinking without active scrolling

downing034:


src/core/components/ChatView.module.css
```diff
@@ -94,11 +94,13 @@
 .assistantContainer {
   margin-top: var(--spacing-sm);
   max-width: 80%;
+  width: 80%;
 }
 
 #progressStream {
   .content {
     color: var(--colors-textTertiary);
+    font-size: 1.5rem;
```
downing034, line 103: shrink the font a bit in the stream a tiny bit (was 1.6) for better alignment and cleaner look

src/core/components/ChatView.module.css
```diff
@@ -94,11 +94,13 @@
 .assistantContainer {
   margin-top: var(--spacing-sm);
   max-width: 80%;
+  width: 80%;
```
downing034, line 97: This keeps the thinking container from starting small and growing as the text is streamed in it to 80%

src/core/components/ChatView.tsx
```diff
@@ -47,7 +47,11 @@ export const ChatView = ({
   const containerRef = useRef<HTMLDivElement>(null);
   const handleContentRendered = useCallback(() => {
     if (containerRef.current) {
-      containerRef.current.scrollTop = containerRef.current.scrollHeight;
+      requestAnimationFrame(() => {
```
downing034, line 50: make sure the page scrolls to the bottom properly when completed

src/core/components/Content.tsx
```diff
@@ -28,22 +29,47 @@ export const ContentComponent = ({
...
+          processedContent =
+            modelConfig.messageProcessors.postProcess(contentToSanitize);
         }
 
         const updatedContent = await sanitizeContent(processedContent);
 
         if (!cancel) {
           setSanitizedContent(updatedContent);
+          if (onRendered) {
+            requestAnimationFrame(() => {
```
downing034, line 58: more scrolling fixes, in this case for cancel

src/core/assets/BrainIcon.tsx
```diff
@@ -0,0 +1,26 @@
+export const BrainIcon = (props: React.SVGProps<SVGSVGElement>) => (
```
samcode206, line 1: would https://lucide.dev/icons/brain work instead? can just 
```import { Brain } from 'lucide-react'``` since we already use that library

samcode206 (approved):

src/core/assets/BrainIcon.tsx
adamrturman: I think the other history PR removes that library and its trash icon.

src/core/components/ChatView.tsx
adamrturman: @downing034 I think you can use `useScrollToBottom` here

src/core/assets/BrainIcon.tsx
downing034: Yeah I was working on avoiding using a library if we don't have to so we're not tied to a upgrades of packages.

src/core/components/ChatView.tsx
downing034: Ahh let me give that a go

adamrturman (approved):

src/core/components/ChatView.tsx
downing034: Locally I added rolled this back to `containerRef.current.scrollTop = containerRef.current.scrollHeight;` and then added `useScrollToBottom(handleContentRendered);` afterwards like our other usage and it continues to scroll back to the top of the ChatView scrollContent container

PR #324, opened 2025-02-11
style: alignment for assistant headers

adamrturman:
## Summary

Sometimes the model will respond with markdown that will get converted into elements like <h3>. We need to apply custom styling in those cases to keep it looking nice.

## Before
Header elements get margin bottom from the global style.

Screenshot 2025-02-11 at 3 49 29 PM

## After
But if its part of the assistant response, then we should give it equal margin top.

Screenshot 2025-02-11 at 3 51 19 PM

### DIFF
--- a/src/core/components/ChatView.module.css
+++ b/src/core/components/ChatView.module.css
@@ -338,3 +338,10 @@ li {
   margin-top: var(--spacing-sm);
   height: var(--spacing-md);
 }
+
+.left h3,
+.left h4,
+.left h5,
+.left h6 {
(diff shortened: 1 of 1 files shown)


nddeluca:
Nice improvement.  Should we apply this to all markdown rendered content and not just .left?  User may enter markdown as well

samcode206 (approved):

adamrturman:
> Nice improvement. Should we apply this to all markdown rendered content and not just .left? User may enter markdown as well

Good idea - will apply to the user messages as well

PR #326, opened 2025-02-11
feat: reasoning Content separation 

samcode206:

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -240,7 +240,7 @@ async function doChat(
 
       if (isContentChunk(chunk)) {
         content += chunk['choices'][0]['delta']['content'];
-
+        // todo: chance to clean up into a helper similar to assembleToolCallFromStream
         switch (modelConfig.id) {
           case 'deepseek-r1': {
             const openTag = '<think>';
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -217,9 +217,7 @@ async function doChat(
       {
         model: id,
         messages: messageHistoryStore.getSnapshot().map((msg) => {
-          // @ts-expect-error reasoningContent may exist
-          if (msg.reasoningContent) {
-            // @ts-expect-error delete from msg
+          if ('reasoningContent' in msg) {
             delete msg.reasoningContent;
           }
           return msg;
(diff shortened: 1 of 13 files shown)


src/core/components/Content.tsx
```diff
@@ -79,24 +53,25 @@ export const ContentComponent = ({
...
     <div data-testid="conversation-message" data-chat-role={role}>
-      {role === 'assistant' && hasThinkingContent && (
-        <ThinkingContent content={content} isStreaming={isRequesting} />
-      )}
       {sanitizedContent !== '' && (
         <span
           className={styles.content}
           dangerouslySetInnerHTML={{ __html: sanitizedContent }}
-        />
+        ></span>
```
adamrturman, line 74: Don't like the single closed tag? lol

samcode206: i do, just copied the old Content component b4 recent changes.

src/core/components/Conversation.tsx
```diff
@@ -54,13 +62,18 @@ const ConversationComponent = ({ messages, onRendered }: ConversationProps) => {
 
         if (message.role === 'assistant' && message.content) {
           return (
-            <AssistantMessage key={index} content={message.content as string} />
+            <AssistantMessage
+              key={index}
+              content={message.content as string}
+              // @ts-expect-error reasoningContent may exist
+              reasoningContent={message.reasoningContent}
```
adamrturman, line 69: I think you could avoid expecting this error if we did some check on the message before passing its possible reasoning content

```
reasoningContent={
  message.role === 'assistant' && 'reasoningContent' in message
    ? message.reasoningContent
    : undefined
}
  ```

File: src/App.tsx [hunk 3]
```diff
@@ -221,24 +230,68 @@ async function doChat(
...
+              );
+            }
+            break;
+          }
+
+          default: {
+            messageStore.setText(content);
+            break;
+          }
+        }
```
adamrturman, line 275: Can you add a comment that it'd be nice to refactor this to a helper similar to `assembleToolCallsFromStream` below? It'll help keep the App lean and it will be easier to test.

adamrturman (approved):

PR #327, opened 2025-02-12
feat: chat history search (title)

adamrturman:
## Summary
- Adds an input to search/filter conversation history by title match
- Moved the conversations state into a useState hook instead of reading directly from localStorage on every render
- Added useMemo for filtering and grouping conversations
- Memoized the HistoryItem component using React.memo
- Memoized the truncateTitle function and its result

## Demo 

https://github.com/user-attachments/assets/aa816cb7-fc56-4e70-abd9-f7a0fec07320

### DIFF
--- a/src/core/components/ChatHistory.tsx
+++ b/src/core/components/ChatHistory.tsx
@@ -90,7 +90,9 @@ export const ChatHistory = ({ setChatHistoryOpen }: ChatHistoryProps) => {
     load();
     // we have to poll local storage
     const id = setInterval(load, 1000);
-    return () => clearInterval(id);
+    return () => {
+      clearInterval(id);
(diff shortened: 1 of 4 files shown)


src/core/components/ChatHistory.tsx
```diff
@@ -135,27 +149,34 @@ export const ChatHistory = ({ setChatHistoryOpen }: ChatHistoryProps) => {
         >
           <div className={styles.newChatButtonAlignment}>
             <NewChatIcon />
+            <input
+              onChange={handleSearchChange}
+              value={searchTerm}
+              placeholder="Search conversations..."
+            />
```
adamrturman, line 156: todo - styling

PR #328, opened 2025-02-12
feat: enable blockchain model on mobile

adamrturman:

### DIFF
--- a/src/core/components/NavBar.module.css
+++ b/src/core/components/NavBar.module.css
@@ -70,8 +70,8 @@
 }
 
 .dropdownMenu {
-  position: absolute;
-  top: 100%;
+  position: fixed;
+  top: 62px;
   left: 50%;
   transform: translateX(-50%);
   background-color: var(--colors-bgSecondary);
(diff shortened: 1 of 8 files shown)


src/core/components/NavBar.module.css
```diff
@@ -70,18 +70,20 @@
 }
 
 .dropdownMenu {
-  position: absolute;
-  top: 100%;
+  position: fixed;
+  top: 62px;
```
adamrturman, line 74: The dropdown menu wasn't actually clickable on mobile - if you selected an option, the click event was happening below the dropdown, triggering close on outside click, making switching on mobile impossible.

src/features/blockchain/services/messages/evm/transfer.ts
```diff
@@ -56,6 +56,13 @@ export class EvmTransferMessage implements ChainMessage<SendToolParams> {
     return this.hasValidWallet ? InProgressTxDisplay : ConnectWalletPrompt;
   }
 
+  private isMobileDevice(): boolean {
+    // Use regex to detect mobile devices
+    return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
+      window.navigator.userAgent,
+    );
```
adamrturman, line 63: This might be a better implementation of `isMobile()`?

PR #341, opened 2025-02-12
fix: adjust click on outside close to encompass the entire dropdown

adamrturman:
## Summary
Clicking the current selection was firing both close on outside click, and _then_ toggling the state back to open.

Move the locator to the parent of the dropdown

## Before

https://github.com/user-attachments/assets/396fe816-6242-4d76-852c-0aae58629f43

## After

https://github.com/user-attachments/assets/c9b6d0ba-03a7-41c9-9d00-26738818eb45

### DIFF
--- a/src/core/components/ModelSelector.tsx
+++ b/src/core/components/ModelSelector.tsx
@@ -1,7 +1,6 @@
 import ModelDropdownButton from './ModelDropdownButton';
 import ModelDropdownExpanded from './ModelDropdownExpanded';
 import { useEffect, useState } from 'react';
-import styles from './NavBar.module.css';
 
 const ModelSelector = () => {
   const [dropdownOpen, setDropdownOpen] = useState(false);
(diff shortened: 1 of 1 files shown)


src/core/components/ModelSelector.tsx
```diff
@@ -1,16 +1,14 @@
 import ModelDropdownButton from './ModelDropdownButton';
 import ModelDropdownExpanded from './ModelDropdownExpanded';
 import { useEffect, useState } from 'react';
-import styles from './NavBar.module.css';
```
adamrturman, line 4: Not used

src/core/components/ModelSelector.tsx
```diff
@@ -1,17 +1,18 @@
...
+import { useEffect, useRef, useState } from 'react';
 
 const ModelSelector = () => {
   const [dropdownOpen, setDropdownOpen] = useState(false);
+  const selectorRef = useRef<HTMLDivElement>(null);
 
   // Handle closing dropdown when clicking outside
   useEffect(() => {
     const handleClickOutside = (event: MouseEvent) => {
-      const dropdown = document.querySelector(`.${styles.dropdownMenu}`);
```
adamrturman, line 12: useRef instead of reaching into the raw DOM

Issue #334, opened 2025-02-12
Take a pass at font sizing


nddeluca:
1rem == 10 px

Adjust global.css for better default font sizes for body, p, h tags, etc
Remove explicitly set font sizes
Remove use of pixel font sizes -- always use rem

When reworking this font-sizing, make sure to open ModelSelector.module.css and remove the font-size in this file search for TODO

PR #342, opened 2025-02-12
feat: new search UI

adamrturman:
## Summary
This sets up the new UI for searching that will eventually support conversation content in addition to titles. This just moves the existing functionality from searching conversation titles in the side bar to a modal. A follow on PR will extend the functionality and adjust the modal UI to show the portion of the content that matches the user search

## Demo 
Modal on desktop
Full overlay on mobile

https://github.com/user-attachments/assets/68f8d5a4-0c89-497f-ba4c-0d84b8d96960


src/core/components/ChatHistory.tsx
```diff
@@ -5,73 +5,27 @@ import {
...
-  return sortedConversations.reduce((groups, conversation) => {
-    const timeGroup = getTimeGroup(conversation.lastSaved);
-    if (!groups[timeGroup]) {
-      groups[timeGroup] = [];
-    }
-    groups[timeGroup].push(conversation);
-    return groups;
-  }, {} as GroupedConversations);
-};
-
```
adamrturman, line 72: Since this functionality now exists in the sidebar and in the search modal, I moved the functions to a helpers file and added unit tests.

src/core/components/SearchModal.module.css
```diff
@@ -0,0 +1,146 @@
+.container {
```
downing034, line 1: We're working on doing the css from a mobile-first approach. Let's change this file to follow similar setup.

Essentially means making all the base classes the mobile stuff and tucking desktop classes inside

```
@media (max-width: 768px) {
  
}
```

src/core/components/SearchModal.module.css
```diff
@@ -0,0 +1,146 @@
...
+}
+
+.timeGroup {
+  margin-bottom: var(--spacing-md);
+}
+
+.timeGroupTitle {
+  margin: var(--spacing-xs) var(--spacing-sm);
+  color: var(--colors-textTertiary);
+  font-size: 12px;
```
downing034, line 112: let's try to avoid setting `font-size` in the css in favor of using global.css font-sizes. If it's difficult to get the design correct with this change, add a TODO here and add this to the [font issue](https://github.com/Kava-Labs/oros/issues/334)

src/core/components/SearchModal.module.css
```diff
@@ -0,0 +1,146 @@
...
+}
+
+.conversationItem:hover {
+  background-color: var(--colors-bgTertiary);
+}
+
+.conversationTitle {
+  color: var(--colors-textPrimary);
+  font-size: 14px;
+  line-height: 20px;
```
downing034, line 132: same as above, let's see if we can use global css here instead of custom overrides

downing034 (approved):

PR #351, opened 2025-02-13
feat: default model selection through query params

samcode206:
## Demo 

https://github.com/user-attachments/assets/d9b4088c-77aa-41f2-8d19-7a4bd5203bb6

### DIFF
--- a/src/core/config/models.ts
+++ b/src/core/config/models.ts
@@ -2,9 +2,7 @@ import {
   blockchainModels,
   SupportedBlockchainModels,
 } from '../../features/blockchain/config/models';
-import {
-  reasoningModels,
-} from '../../features/reasoning/config/models';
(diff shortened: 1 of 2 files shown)


src/core/config/models.ts
```diff
@@ -40,6 +40,20 @@ export const getAllModels = (): ModelConfig[] => {
...
+    return wantedModel;
+  }
+  return null;
+};
+
+export const DEFAULT_MODEL_NAME = defaultModelFromQueryParams()
+  ? defaultModelFromQueryParams()!
+  : import.meta.env.VITE_FEAT_UPDATED_DESIGN
+    ? 'deepseek-r1'
+    : 'gpt-4o';
```
adamrturman, line 59: Should probably check with product for what they'd like displayed in the URL. I would imagine it would be closer to the display names.

PR #355, opened 2025-02-13
feat: search conversation history content

adamrturman:
## Summary
- When modal is open but a user hasn't typed yet
  - show histories like in sidebar (recent descending to past)
  - snippet starts at the first user message
 - When the search is match in the content
   - show the title of the match
   - start the snippet up to 3 words before the match
 - When the search is a match in the title only
   - show the title of the match
   - start the snippet at the first user message
 - When the search isn't present in any content or titles
   - display "No results"     
 
- Matching results are bolded in the title as well as in the content (as long as the match is at least two letters - don't want to match every instance of "e" when starting to type "ethereum" for instance 

## Demo 

https://github.com/user-attachments/assets/13c2f319-3a84-45c6-b269-c951a6bd5e20


e2e/chat.spec.ts
```diff
@@ -384,6 +384,12 @@ describe('chat', () => {
     const chat = new Chat(page);
 
     await page.addInitScript(() => {
+      const systemPrompt = {
+        role: 'system',
+        content:
+          'You are an AI assistant knowledgeable about blockchain technologies.',
+      };
+
```
adamrturman, line 392: Add a system prompt to the mock conversation because we need to ignore that when searching content. Note that it doesn't occur as the match when the user searches "blockchain."

src/core/components/SearchModal.tsx
```diff
@@ -111,17 +113,29 @@ const SearchModal = ({
...
+                          />
+                          <p
+                            data-testid="search-history-content"
+                            className={styles.conversationSnippet}
+                            dangerouslySetInnerHTML={{
+                              __html: highlightMatch(
+                                formatContentSnippet(conversation, searchTerm),
+                                searchTerm,
+                              ),
+                            }}
```
adamrturman, line 137: Allows us to bold the portion of the match

e2e/chat.spec.ts
```diff
@@ -433,21 +480,100 @@ describe('chat', () => {
...
+    await page.addInitScript(() => {
+      localStorage.removeItem('conversations');
+    });
+
+    const searchButton = page.getByTestId('search-conversation-button');
+    await searchButton.click();
+
+    const noResults = page.getByText('No results');
+    await expect(noResults).toBeVisible();
+  });
```
adamrturman, line 578: Verifies that this works for new users or returning users with no chat history

src/core/components/SearchModal.tsx
```diff
@@ -66,9 +68,9 @@ const SearchModal = ({
   return (
     <div className={styles.container}>
       <button
```
downing034, line 70: Is this button just in the PR so the modal can be opened while refactor is in progress? Not I'd expect this to be a layer outside the modal I think so the parent opens the modal.

adamrturman: This component is the button and the modal it launches.

samcode206 (approved):

src/core/components/SearchChatHistory.tsx
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <div className={styles.container}>
+      <button
+        data-testid="search-conversation-button"
+        className={styles.iconButton}
+        onClick={() => setIsOpen(true)}
+        aria-label="Search conversations"
+      >
+        <Search size={20} />
+      </button>
```
adamrturman, line 43: After the redesign, we can plug in the new `ButtonIcon` here and give it the same click handler to open the modal.

PR #356, opened 2025-02-13
refactor: clean up App.tsx move handlers to context 

samcode206:
move all handlers and api calls to context

this is in prep for concurrent conversation support


adamrturman (approved):
Deleting >300 lines from App 😎

PR #366, opened 2025-02-15
feat: reimplement search history

adamrturman:
## Demo
 Open state is preserved between media query changes

https://github.com/user-attachments/assets/bf9d84ec-5962-4cb6-9211-c7b987b46772


src/core/components/SearchChatHistory.tsx
```diff
@@ -1,59 +1,66 @@
...
     <div className={styles.container}>
-      <button
-        data-testid="search-conversation-button"
-        className={styles.iconButton}
-        onClick={() => setIsOpen(true)}
-        aria-label="Search conversations"
+      <SearchChatButton
+        onClick={() => setIsChatHistoryOpen(true)}
+        disabled={!hasConversations}
+      />
```
adamrturman, line 48: Use the existing icon in the component that handles launching the modal.

src/core/components/ModalWrapper.tsx
```diff
@@ -33,7 +34,7 @@ const ModalWrapper = ({
     };
   }, [closeOnOutsideClick, onClose, modalRef]);
 
-  return (
+  return ReactDOM.createPortal(
```
adamrturman, line 37: https://react.dev/reference/react-dom/createPortal#usage

src/core/components/SearchChatHistory.tsx
```diff
@@ -1,53 +1,68 @@
...
   useEffect(() => {
-    if (isOpen) {
+    if (isChatHistoryOpen) {
       // Focus input when modal opens
       inputRef.current?.focus();
+      //  mobile sidebar is closed by default and that contains the component that mounts the search modal,
+      //  but if a user has opened chat history search from a larger screen, then we should update that so search modal
+      //  stays open on screen resize
+      setIsMobileSideBarOpen(true);
+    }
```
adamrturman, line 33: Demo without setting this - it works once the user sets the mobile sidebar open, but not initially.

Not a super likely userflow, but at least its covered


https://github.com/user-attachments/assets/0b1f67df-f001-47df-9168-c31ff6ba28fd

src/core/components/SearchHistory.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { SearchHistoryButton } from '../assets/SearchHistoryButton';
+import { ConversationHistory } from '../context/types';
+
+interface SearchHistoryProps {
+  isSearchHistoryOpen: boolean;
+  setIsSearchHistoryOpen: (i: boolean) => void;
+  setIsMobileSideBarOpen: (i: boolean) => void;
+}
+
+const SearchHistory = ({
```
adamrturman, line 15: This used to be `SearchChatHistory` but I dropped the redundant '-chat-' in the name.

e2e/chat.spec.ts
```diff
@@ -397,7 +397,7 @@ describe('chat', () => {
     await expect(historyEntry).toHaveText('Test Conversation Title');
   });
 
-  test.skip('conversation search functionality', async ({ page }) => {
+  test('conversation search functionality', async ({ page }) => {
```
adamrturman, line 400: Verifies the functionality that worked on the old UI works again (with minor updates)

src/core/components/ChatView.tsx
```diff
@@ -224,10 +224,7 @@ export const ChatView = ({
               </div>
             </div>
 
-            <div
-              className={styles.importantInfo}
-              data-testid="importantInfo"
-            >
```
adamrturman, line 230: Linting issue

src/main.tsx
```diff
@@ -8,9 +8,6 @@ import { createRoot } from 'react-dom/client';
 // Global CSS & Fonts
 //
 import './shared/theme/global.css';
-
-
-
```
adamrturman, line 13: Linting issue

e2e/chat.spec.ts
```diff
@@ -493,7 +493,7 @@ describe('chat', () => {
 
     await chat.goto();
 
-    const searchButton = page.getByTestId('search-conversation-button');
+    const searchButton = page.getByRole('button', { name: 'Search History' });
```
downing034, line 496: I like this type of change. One thing playwright and react testing library do well is they give you selectors you should be able to use without having to use data-testid. Using stuff like getByRole ensures that we've built semantically correct, screen reader friendly code. I'd actually argue in most cases, if you can't use something other than getByTestId then we should actually change how the code is built.

src/core/components/SearchHistory.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const onConversationSelect = (conversationHistory: ConversationHistory) => {
+    loadConversation(conversationHistory);
+    setIsSearchHistoryOpen(false);
+    //  on mobile we want to get the user back to the chat view after selection
+    setIsMobileSideBarOpen(false);
+  };
+
+  return (
+    <div className={styles.container}>
+      <SearchHistoryButton
```
downing034, line 50: I'd like to see this button moved out into the App.tsx components that way whenever we change around the sidebar styling we don't have two components to update and the main chat history component doesn't have to be touched

src/core/components/ChatView.tsx
```diff
@@ -224,7 +224,7 @@ export const ChatView = ({
               </div>
             </div>
 
-            <div className={styles.importantInfo} data-testid="importantInfo">
```
adamrturman: Will do a follow-on sometime this week to see how many more of these can be removed.

downing034 (approved):

samcode206 (approved):

PR #368, opened 2025-02-17
fix: search modal close icon

adamrturman:
## Summary
- Desktop: styled similarly to chat gpt, the desktop search history modal has a close icon as an "adornment button" styled inside the input to save on screen real estate.
- Mobile: refactor the existing design to use the `ButtonIcon` component & styles

Mobile design leverages the close icon (header style) that is built into the modal wrapper.

## Demo 

https://github.com/user-attachments/assets/730e1b8f-fd7d-47ca-ab73-2183846aaefa


src/core/components/SearchChatHistory.module.css
```diff
@@ -15,10 +14,6 @@
   transition: color 0.2s ease;
 }
 
-.iconButton:hover {
-  color: var(--colors-textMuted);
-}
-
```
adamrturman, line 21: Don't need to explicitly change color on hover now that we've got base ButtonIcon styles 🦾

src/core/components/SearchChatHistory.module.css
```diff
@@ -68,6 +64,7 @@
   border-radius: var(--borderRadius-sm);
   padding: 0 var(--spacing-md);
   color: var(--colors-textPrimary);
+  padding-right: calc(var(--spacing-md) * 3);
```
adamrturman, line 67: We want to style the close button inside the input, but we need enough padding so that on hover, there is no collision with the text.

samcode206 (approved):

samcode206 (approved):

PR #369, opened 2025-02-18
feat: new chat default title

adamrturman:
## Summary
- when a new chat is started, it is added to history with a placeholder title (New Chat)
- when a user engages with the chat and gets a response, it gets the AI-summarization as previous
- if a user starts multiple new chats or switches models without starting a chat, the previous empty chat is deleted from history

### DIFF
--- a/src/core/context/useMessageSaver.test.ts
+++ b/src/core/context/useMessageSaver.test.ts
@@ -52,6 +52,162 @@ describe('useMessageSaver', () => {
     vi.clearAllMocks();
   });
 
+  it('should not save conversation with less than 3 messages', () => {
+    mockMessageHistoryStore.getSnapshot.mockReturnValue([
+      { role: 'user', content: 'Hello' },
+      { role: 'assistant', content: 'Hi' },
+    ]);
+
+    renderHook(() =>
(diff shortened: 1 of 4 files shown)


src/core/context/useMessageSaver.tsx
```diff
@@ -16,15 +18,41 @@ export const useMessageSaver = (
...
+        !allConversations[id]
+      ) {
+        // Clean up any existing empty placeholder chats
+        Object.entries(allConversations).forEach(([convId, conv]) => {
+          if (
+            conv.title === PLACEHOLDER_CHAT_TITLE &&
+            conv.conversation.length === 1
+          ) {
+            delete allConversations[convId];
+          }
```
adamrturman, line 38: Prevents history from being overloaded with empty entries if a user clicks the new chat button repeated or switches between models.

src/core/context/useMessageSaver.test.ts
```diff
@@ -91,7 +91,6 @@ describe('useMessageSaver', () => {
       ),
     );
 
-    // Trigger the onChange callback
```
adamrturman, line 94: todo - add comments back

PR #370, opened 2025-02-18
fix: add max scale to prevent zoom on inputs

nddeluca:
Without the maximum-scale set, the edit history and search input on mobile will zoom in on click, but not zoom out after typing.  This change keeps the scale the same.

Tested on iphone 15 simulator and fixes the zoom behavior.


nddeluca:
This may be best as a temporary fix -- the other fix is to use 16px or above for all inputs and apply that to the edit title and search inputs.

adamrturman (approved):

PR #372, opened 2025-02-18
feat: new chat placeholder applied on first user message

adamrturman:
## Summary

- "New Chat" is the history title placeholder when a user sends the first message.
-  It persists as that through the assistant streaming response 
- When the assistant's reply is finished, we do the AI summarization as previous

## Demo 

https://github.com/user-attachments/assets/58088a68-8be2-470c-a970-3e8feea08d73

### DIFF
--- a/e2e/chat.spec.ts
+++ b/e2e/chat.spec.ts
@@ -287,6 +287,15 @@ describe('chat', () => {
     await chat.waitForStreamToFinish();
     await chat.waitForAssistantResponse();
 
+    //  we don't want to open a new conversation before the title generation completes
+    await page.waitForFunction(() => {
+      const historyEntry = document.querySelector(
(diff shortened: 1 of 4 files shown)


e2e/chat.spec.ts
```diff
@@ -287,6 +287,15 @@ describe('chat', () => {
...
 
+    //  we don't want to open a new conversation before the title generation completes
+    await page.waitForFunction(() => {
+      const historyEntry = document.querySelector(
+        '[data-testid="chat-history-entry"]:first-child',
+      );
+      const title = historyEntry?.textContent;
+      return title && title !== 'New Chat';
+    });
+
```
adamrturman, line 298: I think we should be able to remove this once #375 is merged in

adamrturman:
Closing in favor of #382

PR #374, opened 2025-02-18
fix: edit history title - UI first, then storage

adamrturman:
## Summary
Updating local storage is synchronous and for users with long histories, it can cause the UI to lag when editing.

Let's update the UI immediately and then fire an async request to update storage

## Before

https://github.com/user-attachments/assets/8fa071dc-cd9f-48b0-8387-b48a03eb82f6

## After

https://github.com/user-attachments/assets/d40bf9ea-e3a1-455f-b94f-7e91f84bd5a6

### DIFF
--- a/src/core/components/ChatHistory.tsx
+++ b/src/core/components/ChatHistory.tsx
@@ -14,6 +14,7 @@ import { useAppContext } from '../context/useAppContext';
 import { EllipsisVertical, Bot, Pencil, Trash2, X } from 'lucide-react';
 import { groupConversationsByTime } from '../utils/conversation/helpers';
 import ButtonIcon from './ButtonIcon';
+import _ from 'lodash';
 
 interface ChatHistoryProps {
   onHistoryItemClick: Dispatch<SetStateAction<boolean>>;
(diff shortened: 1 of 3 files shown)


src/core/components/ChatHistory.tsx
```diff
@@ -158,18 +159,29 @@ const HistoryItem = memo(
       const trimmedTitle = newTitle.trim();
       if (trimmedTitle === '') {
         setNewTitle(title);
-      } else {
-        saveToLocalStorage(trimmedTitle);
-        setNewTitle(trimmedTitle);
```
adamrturman, line 163: This was the root of the issue. Updating local storage is synchronous, so if we do this at the same time as updating the UI, we can see a significant lag for users with long histories.

https://github.com/user-attachments/assets/ada77df1-016c-4b8b-88cb-f8dbd530bbc8

src/core/components/ChatHistory.tsx
```diff
@@ -158,18 +158,28 @@ const HistoryItem = memo(
...
+        return;
```
adamrturman, line 161: Early return instead of a long else block

src/core/components/ChatHistory.tsx
```diff
@@ -245,17 +255,13 @@ const HistoryItem = memo(
                 }}
               />
             ) : (
-              <small data-testid="chat-history-entry">{title}</small>
+              <small data-testid="chat-history-entry">{newTitle}</small>
```
adamrturman, line 258: We should get the most recent version of the title.

src/core/components/ChatHistory.tsx
```diff
@@ -158,11 +158,23 @@ const HistoryItem = memo(
...
         setNewTitle(title);
-      } else {
-        saveToLocalStorage(trimmedTitle);
-        setNewTitle(trimmedTitle);
+        return;
       }
+
+      //  Update UI first
+      setNewTitle(trimmedTitle);
       setEditingTitle(false);
```
adamrturman, line 166: I made a slight UX change here. Once the user saves the new title, we exit them out of editing mode

downing034 (approved):

PR #379, opened 2025-02-19
fix: delete from UI, then from local storage

adamrturman:
## Summary
This is similar to the change recently made for title editing. If a user clicks to delete a chat, we don't want to wait for a synchronous action to complete before updating the UI.

We will track conversations in local state and remove it from there first and then schedule the update to local storage.

## Before

https://github.com/user-attachments/assets/5dbade08-8683-4c8d-8ec1-c695642843c7

## After

https://github.com/user-attachments/assets/e66add22-5801-4331-b929-ea3c52c6ecf3

### DIFF
--- a/src/core/components/ChatHistory.tsx
+++ b/src/core/components/ChatHistory.tsx
@@ -145,12 +145,12 @@ interface HistoryItemProps {
 
 const HistoryItem = memo(
   ({
-    conversation,
-    handleChatHistoryClick,
-    deleteConversation,
-    isMenuOpen,
-    onMenuClick,
-  }: HistoryItemProps) => {
+     conversation,
+     handleChatHistoryClick,
+     deleteConversation,
+     isMenuOpen,
+     onMenuClick,
+   }: HistoryItemProps) => {
     const { id, title } = conversation;
     const { conversationID } = useAppContext();
     const isSelected = conversationID === id;
--- a/src/core/components/ChatHistory.tsx
+++ b/src/core/components/ChatHistory.tsx
@@ -28,23 +28,56 @@ export const ChatHistory = ({ onHistoryItemClick }: ChatHistoryProps) => {
     hasConversations,
   } = useAppContext();
 
+  const conversationsToRecord = (convs: ConversationHistory[]) => {
+    const record: Record<string, ConversationHistory> = {};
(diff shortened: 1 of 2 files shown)


src/core/components/ChatHistory.tsx
```diff
@@ -38,19 +38,26 @@ export const ChatHistory = ({
...
-      if (
-        allConversations[id] &&
-        id === messageHistoryStore.getConversationID()
-      ) {
+      if (id === messageHistoryStore.getConversationID()) {
         startNewChat();
       }
 
-      delete allConversations[id];
-      localStorage.setItem('conversations', JSON.stringify(allConversations));
```
adamrturman, line 53: 1. Update the UI with `startNewChat()`
2. Schedule all of the storage updates to happen after

src/core/components/ChatHistory.tsx
```diff
@@ -24,33 +24,49 @@ export const ChatHistory = ({
...
-        id === messageHistoryStore.getConversationID()
-      ) {
+      // First, update the UI by removing the conversation from local state
+      setLocalConversations((prev) => {
+        const updated = { ...prev };
+        delete updated[id];
+        return updated;
+      });
+
+      if (id === messageHistoryStore.getConversationID()) {
```
samcode206, line 49: this is now conversationID you can grab it from useAppContext

samcode206:
```
  const { conversationID } = useAppContext();
```

samcode206 (approved):

nddeluca (approved):
Nice improvement to the UI responsiveness 🚀 

One 🐛  I found is the `conversations` empty state rendering doesn't trigger based on the local conversations, so that UX update is delayed until after the local storage is updated.  Not blocking though, can be improved here or as a follow on

adamrturman:
> Nice improvement to the UI responsiveness 🚀 
> 
> 
> 
> One 🐛  I found is the `conversations` empty state rendering doesn't trigger based on the local conversations, so that UX update is delayed until after the local storage is updated.  Not blocking though, can be improved here or as a follow on

Thanks @nddeluca - good find on that regression on the empty history placeholder. I'll add a fix and a test to verify.

PR #382, opened 2025-02-19
feat: new chats start with a placeholder

adamrturman:

### DIFF
--- a/src/core/components/ChatHistory.tsx
+++ b/src/core/components/ChatHistory.tsx
@@ -160,10 +160,6 @@ const HistoryItem = memo(
         setDidEditTitle(true);
       }
 
-      if (trimmedTitle !== title) {
-        setDidEditTitle(true);
-      }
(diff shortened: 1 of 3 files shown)


src/core/components/ChatHistory.tsx
```diff
@@ -229,13 +235,19 @@ const HistoryItem = memo(
 
     useEffect(() => {
       if (editingTitle) {
+        const conversations = JSON.parse(
+          localStorage.getItem('conversations') ?? '{}',
+        );
+        const currentTitle = conversations[id]?.title ?? title;
+        setNewTitle(currentTitle);
+
```
adamrturman, line 243: The transition of the title from "New Chat" ➡️ "Summarization" won't trigger any react re-render, so we should double check local storage here to be sure we have the most current value for the title

samcode206 (approved):

PR #385, opened 2025-02-20
feat: blockchain model context window (gpt-tokenizer)

adamrturman:
## Summary
- uses [gpt-tokenizer](https://github.com/niieani/gpt-tokenizer) to encode the conversation into tokens and approximate the total usage so we can warn users to switch to a new conversation as they approach the limit


src/core/utils/conversation/helpers.ts
```diff
@@ -189,3 +192,37 @@ const removeInitialSystemMessage = (conversation: ConversationHistory) => {
...
+ * Pure function to calculate context metrics based on messages and max tokens
+ * @param messages Array of chat messages
+ * @returns ContextMetrics object
+ */
+export async function calculateContextMetrics(
+  chatMessages: ChatMessage[],
+): Promise<ContextMetrics> {
+  const messages = chatMessages as TextChatMessage[];
+  const maxTokens = MAX_TOKENS;
+  const tokensUsed = encodeChat(messages, 'gpt-4o').length;
```
adamrturman, line 209: We can encode the entire chat at once, so long as its one of the string literals it expects.

src/core/context/types.ts
```diff
@@ -56,3 +56,8 @@ export type AppContextType = {
   conversations: ConversationHistory[];
   hasConversations: boolean;
 };
+
+export interface TextChatMessage {
+  role: 'system' | 'assistant' | 'user';
+  content: string;
+}
```
adamrturman, line 63: We have to do a bit of massaging since the 3rd party library doesn't like getting passed exact types like `import type { ChatCompletionMessageParam } from 'openai/resources/index';
`

src/core/components/ChatView.tsx
```diff
@@ -7,6 +7,8 @@ import { useAppContext } from '../context/useAppContext';
 import { NavBar } from './NavBar';
 import type { ChatMessage } from '../stores/messageHistoryStore';
 import { useMessageHistory } from '../hooks/useMessageHistory';
+import { isWithinTokenLimit } from 'gpt-tokenizer';
```
adamrturman, line 10: This is a very useful method that they expose that is ideal for the use case of disabling the input if a user attempts to input a ton of text that itself would overwhelm the context window.

e2e/chat.spec.ts
```diff
@@ -700,4 +700,20 @@ describe('chat', () => {
...
+
+    await chat.switchToBlockchainModel();
+
+    //  Generate 1 million tokens
+    await page.getByTestId('chat-view-input').fill('Token '.repeat(10 ** 6));
+
+    const sendButton = page.getByRole('button', { name: 'Send Chat' });
+
+    await expect(sendButton).toBeDisabled();
+  });
```
adamrturman, line 718: Just for development purposes - we can verify this via unit tests

src/core/context/AppContextProvider.tsx
```diff
@@ -543,12 +546,39 @@ async function syncWithLocalStorage(
...
+      const metrics = await modelConfig.contextLimitMonitor(
+        messages,
+        contextLength,
+      );
+      tokensRemaining = metrics.tokensRemaining;
+    } catch {
+      //  estimate the amount
+    }
+  }
+
```
adamrturman, line 574: This will get simplified after this when I rebase the deepseek branch and both models have a `contextLimitMonitor` function. 

We will call the model specific function and if it fails, we'll do a rough estimate based on character length

src/core/context/AppContextProvider.tsx
```diff
@@ -230,7 +230,7 @@ export const AppContextProvider = (props: {
       // save to local storage (pre-request with user's prompt)
       syncWithLocalStorage(
         conversationID,
-        modelConfig.id,
+        modelConfig,
```
adamrturman, line 233: We'll need both the id and the context length

src/core/context/AppContextProvider.tsx
```diff
@@ -528,10 +528,13 @@ async function callTools(
 
 async function syncWithLocalStorage(
   conversationID: string,
-  modelID: string,
+  modelConfig: ModelConfig,
   messageHistoryStore: MessageHistoryStore,
   client: OpenAI,
+  //  todo - reimplement with deepseek
+  // apiResponse?: ChatCompletionChunk, // Optional response for token tracking
```
adamrturman, line 535: This PR shows a preview of how this will work: https://github.com/Kava-Labs/oros/blob/92220c8ad70acba29338b2ba6bac89234489e30b/src/core/context/AppContextProvider.tsx#L435

src/core/context/AppContextProvider.tsx
```diff
@@ -543,12 +546,39 @@ async function syncWithLocalStorage(
...
+
+  const isDeepseekModel = id.includes('deepseek');
+  const isGPTModel = id.includes('gpt');
+
+  //  todo - refactor to use modelConfig for both and simplify conditionals
+  if (isDeepseekModel) {
+    //  If we get a response with usage, use that to calculate
+  } else {
+    //  todo - plug in deepseek approximation function
+  }
```
adamrturman, line 560: This PR shows a preview of how this will work: https://github.com/Kava-Labs/oros/blob/92220c8ad70acba29338b2ba6bac89234489e30b/src/core/context/AppContextProvider.tsx#L435

samcode206 (approved):

PR #386, opened 2025-02-21
feat: blockchain model context window (js-tiktoken)

adamrturman:
## Summary
- uses [js-tiktoken](https://github.com/dqbd/tiktoken/tree/main/js) to encode the conversation into tokens and approximate the total usage so we can warn users to switch to a new conversation as they approach the limit


adamrturman:
Closing in favor of #385

PR #388, opened 2025-02-24
feat: input adornment message (context limit)

adamrturman:
## Summary
- adds a (currently unimplemented) UI treatment that affixes a message to the top of the input. 
- the user can close this message
- the user can start a new chat 

## Demo 
- Works across screen sizes
- closes on icon click
 
https://github.com/user-attachments/assets/3a8a98bd-62c2-4476-ba26-93b2633b731b

- start new chat

https://github.com/user-attachments/assets/93d251db-e6d7-4efc-b716-394a6149c82c


src/core/components/ChatInput.tsx
```diff
@@ -16,6 +17,8 @@ const ChatInput = ({
   onCancel,
   setShouldAutoScroll,
 }: ChatInputProps) => {
+  const [showInputAdornmentMessage, setShowInputAdornmentMessage] =
```
adamrturman, line 20: I imagine this state will eventually be managed in a hook that integrates with the utility to check the context limit. We could also display custom text based on how close the user is to going over the limit as well as other UI treatments, like disabling the send chat button.

src/shared/theme/themes.ts
```diff
@@ -84,6 +85,7 @@ export const baseTheme: Theme = {
     accentTransparent: 'rgba(255, 67, 62, 0.75)',
     accentTwo: 'rgb(147, 130, 215)',
     accentTwoTransparent: 'rgba(147, 130, 215, 0.75)',
+    accentTwoMuted: 'rgba(147, 130, 215, 0.4)',
```
adamrturman, line 88: Neither of the other opacities looked quite right

samcode206 (approved):

PR #390, opened 2025-02-24
feat: File/image upload with document previews

samcode206:
## Demo 

https://github.com/user-attachments/assets/baf83178-b459-407f-b2e5-842558bf1656


src/core/components/ButtonIcon.tsx
```diff
@@ -49,10 +50,20 @@ const ButtonIcon = ({
...
     : undefined;
 
+  const tooltipStyle = tooltipDelay
+    ? {
+        transitionDelay: `${tooltipDelay}ms`,
+      }
+    : backgroundColorOveride
+      ? {
+          backgroundColor: `${backgroundColorOveride}`,
+        }
```
adamrturman, line 64: We may need to set a custom background to make the tooltip contrast with the element behind it.

src/core/context/utils.ts
```diff
@@ -27,6 +33,27 @@ export const newConversation = () => {
   };
 };
 
+const analyzeImage = async (
+  client: OpenAI,
+  msg: ChatCompletionMessageParam,
+) => {
+  const data = await client.chat.completions.create({
+    model: 'qwen2.5-vl-7b-instruct',
+    messages: [msg], // todo: add system prompt for qwen
```
adamrturman, line 42: Is this todo completed?

adamrturman (approved):

PR #392, opened 2025-02-25
feat: context limit (deepseek)

adamrturman:
## Summary
- as a user engages in a conversation with the reasoning model, we track the token usage from the api response on that conversation's history so we can eventually warn the user and provide good UX as they approach the limit
- adjusts the chat input component to check for available context and compare the input to determine whether to enable the send button


src/core/context/AppContextProvider.tsx
```diff
@@ -543,12 +573,32 @@ async function syncWithLocalStorage(
...
+  } else {
+    //  Fallback to estimation if no token info available
+    const estimatedUsage =
+      //  todo - move this utility to the model config
+      modelConfig.id === 'deepseek-r1'
+        ? estimateTokenUsage(messages)
+        : //  this will be replaced by a gpt-specific utility
+          { totalTokens: 0 };
+    tokensRemaining = Math.max(0, contextLength - estimatedUsage.totalTokens);
+  }
```
adamrturman, line 593: This will eventually become something like `modelConfig.contextLimitMonitor(messages)` and then the model config will specify which function to run.

src/features/reasoning/helpers.test.ts
```diff
@@ -0,0 +1,251 @@
...
+    ];
+
+    const result = estimateTokenUsage(messages);
+
+    // "Hello, how are you?" is 19 chars, so ~5 tokens
+    // "I am doing well, thank you for asking!" is 42 chars, so ~10 tokens
+    expect(result.promptTokens).toBe(5);
+    expect(result.completionTokens).toBe(10);
+    expect(result.totalTokens).toBe(15);
+  });
```
adamrturman, line 29: Unfortunately, this is the best we will be able to do for restricting a user from sending a message that might overwhelm the context limit. We only have access to deepseek usage _after_ the request, so we'll have to use this as a guard on the input for the reasoning model.

src/core/components/ChatInput.tsx
```diff
@@ -102,6 +103,16 @@ const ChatInput = ({
     }
   }, []);
 
+  const allConversations: Record<string, ConversationHistory> = JSON.parse(
+    localStorage.getItem('conversations') ?? '{}',
+  );
+
+  const currentConversation = allConversations[conversationID];
```
adamrturman, line 110: I've seen a few repetitions of this in various places - might make sense to simplify and expose on app context?

src/core/components/InputAdornmentMessage.tsx
```diff
@@ -4,12 +4,12 @@ import ButtonIcon from './ButtonIcon';
 import { X as CloseX } from 'lucide-react';
 import { useAppContext } from '../context/useAppContext';
 
-interface NewChatPromptProps {
+interface InputAdornmentMessageProps {
```
adamrturman, line 7: Renamed to match component

src/core/utils/conversation/hasSufficientRemainingTokens.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  describe('blockchain model', () => {
+    it('calls isWithinTokenLimit with correct arguments', () => {
+      hasSufficientRemainingTokens('gpt-4o', 'test input', 2000);
+
+      expect(isWithinTokenLimit).toHaveBeenCalledWith('test input', 2000);
+      expect(estimateTokenUsage).not.toHaveBeenCalled();
+    });
+  });
```
adamrturman, line 31: This function simply routes between two helpers, one from a third party app and one of ours that has its own unit tests.

src/core/components/ChatInput.tsx
```diff
@@ -131,7 +142,11 @@ const ChatInput = ({
             aria-label="Send Chat"
             disabled={
               (!isRequesting && inputValue.length === 0) ||
-              !isWithinTokenLimit(inputValue, modelConfig.contextLength)
+              !hasSufficientRemainingTokens(
+                modelConfig.id,
+                inputValue,
+                remainingContextWindow,
+              )
```
adamrturman, line 149: - For gpt, we have a 3rd party hook that checks 
- For deepseek, we have to do a rough estimate by characters since we only know token usage _after_ the model responds

samcode206 (approved):

src/features/reasoning/helpers.ts
```diff
@@ -0,0 +1,68 @@
...
+  messages: ChatMessage[],
+): {
+  totalTokens: number;
+} => {
+  // Very rough token estimation (about 4 chars per token)
+  let promptTokens = 0;
+  let completionTokens = 0;
+
+  messages.forEach((msg) => {
+    const content = typeof msg.content === 'string' ? msg.content : '';
```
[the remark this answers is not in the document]
adamrturman: Ok - I can add in an overhead estimate per message of ~3-5 tokens.

PR #394, opened 2025-02-25
fix: add method to wait for summarization

adamrturman:
## Summary

I've noticed some flakiness in this test, especially around capturing only the beginning portion of the assistant's response.

I think this is because the test is proceeding when it gets a 200 response from the chat completions endpoint, but that can actually be the request to get the chat title summary instead.

This specifies which model the request was sent to (summarizations go to the leaner model, chats go to the full one) and uses that to wait for both of those to complete before proceeding

### DIFF
--- a/e2e/chat.spec.ts
+++ b/e2e/chat.spec.ts
@@ -292,6 +292,15 @@ describe('chat', () => {
     //  Switching between conversation histories
     let blockchainQuestionConversation =
       await chat.getMessageElementsWithContent();
+
+    console.log('0', await blockchainQuestionConversation[0].innerText());
+    console.log('1', await blockchainQuestionConversation[1].innerText());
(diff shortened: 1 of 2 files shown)


e2e/chat.spec.ts
```diff
@@ -285,11 +286,21 @@ describe('chat', () => {
...
+    await chat.waitForSummarizationToFinish();
     await chat.waitForAssistantResponse();
 
     //  Switching between conversation histories
     let blockchainQuestionConversation =
       await chat.getMessageElementsWithContent();
+
+    console.log('0', await blockchainQuestionConversation[0].innerText());
+    console.log('1', await blockchainQuestionConversation[1].innerText());
+    console.log('2', await blockchainQuestionConversation[2].innerText());
```
adamrturman, line 298:
```suggestion
```

adamrturman:
Closing in favor of #397

PR #398, opened 2025-02-26
fix: don't rename an existing conversation

adamrturman:
## Before
Gets renamed to the placeholder 

https://github.com/user-attachments/assets/8cffdda6-ad28-4a78-bb9d-30825fed1e01

## After

https://github.com/user-attachments/assets/1e18e77f-36c8-4a43-9922-833d43e10629


e2e/Chat.ts
```diff
@@ -1,4 +1,4 @@
-import { ElementHandle, expect, Locator, Page } from '@playwright/test';
```
adamrturman, line 1: Unused (was previously removed,  but reintroduced in this revert #396

PR #399, opened 2025-02-26
feat: show context limit warning UX (demo)

adamrturman:
## Summary
- when a user crosses a threshold of token usage for a conversation, the warning is shown
- the user can close the warning and it won't be shown again while that conversation remains in view
- if the user returns to that conversation later, the warning will be shown (we aren't storing that preference on the conversation history)

## Demo 

https://github.com/user-attachments/assets/dec4e6e3-0259-4cb9-9c32-797218c0269c

## Limit Reached
- Warning shown
- no close icon (just start new chat)
- send button disabled

Screenshot 2025-02-27 at 12 50 42 PM

### DIFF
--- a/src/core/components/InputAdornmentMessage.module.css
+++ b/src/core/components/InputAdornmentMessage.module.css
@@ -1,29 +1,24 @@
-.cautionBanner {
+.banner {
   display: flex;
   align-items: center;
   gap: var(--spacing-sm);
   padding: var(--spacing-sm);
   width: 98%;
-  background-color: var(--colors-accentTwoMuted);
   font-size: 1.1rem;
   color: var(--colors-textPrimary);
   margin-bottom: -1rem;
   border-radius: var(--borderRadius-sm);
+}
+
+.cautionBanner {
+  composes: banner;
+  background-color: var(--colors-accentTwoMuted);
 }
 
 .warningBanner {
-  display: flex;
-  align-items: center;
-  gap: var(--spacing-sm);
-  padding: var(--spacing-sm);
-  width: 98%;
+  composes: banner;
   background-color: var(--colors-accentMuted);
-  font-size: 1.1rem;
-  color: var(--colors-textPrimary);
-  margin-bottom: -1rem;
-  border-radius: var(--borderRadius-sm);
 }
-
 
 .message {
   flex: 1;
--- a/src/core/components/InputAdornmentMessage.module.css
+++ b/src/core/components/InputAdornmentMessage.module.css
@@ -1,4 +1,4 @@
-.banner {
+.cautionBanner {
   display: flex;
   align-items: center;
   gap: var(--spacing-sm);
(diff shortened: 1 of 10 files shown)


samcode206 (approved):

src/shared/theme/themes.ts
```diff
@@ -83,6 +84,7 @@ export const baseTheme: Theme = {
     textMuted: 'rgb(150, 150, 150)',
     accent: 'rgb(255, 67, 62)',
     accentTransparent: 'rgba(255, 67, 62, 0.75)',
+    accentMuted: 'rgba(255, 67, 62, 0.4)',
```
adamrturman, line 87: `accentTransparent` was a little too jarring, so I applied the same `0.4` opacity as I did with the `accentTwoMuted` 

https://github.com/Kava-Labs/oros/pull/388/files#r1968356748

src/core/types/models.ts
```diff
@@ -18,6 +18,8 @@ export interface BaseModelConfig {
     messages: ChatMessage[],
     contextLimit: number,
   ) => Promise<ContextMetrics>;
+  contextWarningThresholdPercentage: number;
+  conversationResetThreshold: number;
```
adamrturman, line 22: Adding these as model-specific in case we want to adjust limits more specifically

src/core/components/InputAdornmentMessage.module.css
```diff
@@ -4,13 +4,22 @@
...
+.cautionBanner {
+  composes: banner;
+  background-color: var(--colors-accentTwoMuted);
+}
+
+.warningBanner {
+  composes: banner;
+  background-color: var(--colors-accentMuted);
+}
+
```
adamrturman, line 22: Diff here is a little weird

- the existing `banner` class is no longer used directly in the component
- it acts as a base class for the two new ones: caution & warning
- the only difference between those two is the background color, so this keeps the CSS DRY'er

src/core/components/InputAdornmentMessage.tsx
```diff
@@ -34,8 +43,9 @@ export const InputAdornmentMessage: React.FC<NewChatPromptProps> = ({
           icon={CloseX}
           size={12}
           aria-label="Close New Chat Prompt"
-          className={styles.closeIcon}
-          onClick={() => setShowInputAdornmentMessage(false)}
+          className={closeIconStyle}
+          onClick={onCloseClick}
+          disabled={shouldDisableChat}
```
adamrturman, line 48: When the button is not visible, be sure that its not active.

src/core/utils/conversation/helpers.ts
```diff
@@ -196,7 +196,7 @@ const removeInitialSystemMessage = (conversation: ConversationHistory) => {
 /**
  *
  * Function to calculate context metrics based on messages and model's context length
- * @param messages Array of chat messages
+ * @param chatMessages  Array of chat messages
```
adamrturman, line 199: Renamed to match parameter (which was renamed before but not updated here)

src/core/types/models.ts
```diff
@@ -18,6 +18,8 @@ export interface BaseModelConfig {
     messages: ChatMessage[],
     contextLimit: number,
   ) => Promise<ContextMetrics>;
+  contextWarningThresholdPercentage: number;
+  conversationResetTokenThreshold: number;
```
adamrturman, line 22: Adding these as model-specific in case we want to adjust limits more specifically

src/shared/theme/themes.ts
downing034: I still think the orange warning may. be a bit jarring. Might be worth swapping in the purple and adding a screenshot to get feedback from Kevin.

samcode206 (approved):

PR #401, opened 2025-02-27
fix: reset system prompt on model change

downing034:
## Summary
- check and reset message history with system prompt when model change

### DIFF
--- a/src/core/context/AppContextProvider.tsx
+++ b/src/core/context/AppContextProvider.tsx
@@ -146,6 +146,30 @@ export const AppContextProvider = (props: {
     [conversations],
   );
 
+  const ensureCorrectSystemPrompt = useCallback(
+    (messageStore: MessageHistoryStore, modelConfig: ModelConfig) => {
+      const messages = messageStore.getSnapshot();
+      const systemPrompt = modelConfig.systemPrompt;
+
+      // If no messages, add the system message
+      if (messages.length === 0) {
+        messageStore.addMessage({
+          role: 'system',
+          content: systemPrompt,
+        });
+      } else if (messages[0].role === 'system') {
+        // If first message is system, update it
+        const updatedMessages = [...messages];
+        updatedMessages[0] = {
+          role: 'system',
+          content: systemPrompt,
+        };
+        messageStore.setMessages(updatedMessages);
+      }
+    },
+    [],
+  );
+
   const handleModelChange = useCallback(
     (modelName: SupportedModels) => {
       const newConfig = getModelConfig(modelName);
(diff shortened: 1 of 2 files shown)


src/core/context/AppContextProvider.tsx
```diff
@@ -303,13 +335,6 @@ export const AppContextProvider = (props: {
     }
   }, [toolCallStreamStore]);
 
-  const handleReset = useCallback(() => {
```
downing034, line 306: This was not being used anymore (used to be the old trash icon before chat history's handler). Removed it to avoid confusion around systemPrompt reset

src/core/context/AppContextProvider.tsx
```diff
@@ -146,10 +146,45 @@ export const AppContextProvider = (props: {
     [conversations],
   );
 
-  const handleModelChange = useCallback((modelName: SupportedModels) => {
-    const newConfig = getModelConfig(modelName);
-    setModelConfig(newConfig);
-  }, []);
+  const ensureCorrectSystemPrompt = useCallback(
```
downing034, line 149: create a generalized DRY reset system prompt function

src/core/context/AppContextProvider.tsx
```diff
@@ -194,13 +231,8 @@ export const AppContextProvider = (props: {
...
   useEffect(() => {
-    if (!messageHistoryStore.getSnapshot().length) {
-      messageHistoryStore.addMessage({
-        role: 'system' as const,
-        content: modelConfig.systemPrompt,
-      });
-    }
-  }, [messageHistoryStore, modelConfig.systemPrompt]);
+    ensureCorrectSystemPrompt(messageHistoryStore, modelConfig);
+  }, [messageHistoryStore, modelConfig, ensureCorrectSystemPrompt]);
```
downing034, line 235: I can't decide if this useEffect is needed anymore. The only times we should be resetting and adding the systemPrompt in is on new chat started and on model change which are both covered above. Makes me think this is a redundant update. Thoughts?

adamrturman: Before seeing this comment, I was thinking "Do we still need this?"

I did some manual testing with it removed and it seems good.

src/core/context/AppContextProvider.tsx
```diff
@@ -146,10 +146,45 @@ export const AppContextProvider = (props: {
...
+      } else if (messages[0].role === 'system') {
+        // If first message is system, update it
+        const updatedMessages = [...messages];
+        updatedMessages[0] = {
+          role: 'system',
+          content: systemPrompt,
+        };
+        messageStore.setMessages(updatedMessages);
+      }
+    },
```
adamrturman, line 169: Would it be overkill to also check for `messages.length === 1` here? Shouldn't we only ever be resetting the system prompt (and then the messageStore) if its empty or only has a system prompt? Spreading existing messages after an updated system prompt wouldn't actually happen, right?

```
} else if (messages[0].role === 'system' && messages.length === 1) {
  const updatedMessages = [ {
    role: 'system',
    content: systemPrompt,
  }];
  messageStore.setMessages(updatedMessages);
}
      ```

adamrturman (approved):

PR #404, opened 2025-02-28
refactor: adjust `ContextMetrics` properties

adamrturman:
## Summary
We only need to store `tokensRemaining` on ChatHistory in local storage and can derive any other information from the model config. This removes these unused properties on the `ContextMetrics` type

### DIFF
--- a/src/core/types/models.ts
+++ b/src/core/types/models.ts
@@ -65,7 +65,5 @@ export interface ModelRegistry {
 }
 
 export interface ContextMetrics {
-  tokensUsed: number;
   tokensRemaining: number;
-  percentageRemaining: number;
 }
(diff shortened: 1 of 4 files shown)


src/core/types/models.ts
```diff
@@ -65,7 +65,5 @@ export interface ModelRegistry {
 }
 
 export interface ContextMetrics {
-  tokensUsed: number;
   tokensRemaining: number;
-  percentageRemaining: number;
```
adamrturman, line 70: Since we have `tokensRemaining` in local storage and max tokens on the model config, we can calculate percentage remaining when we need to, rather than adding to conversation history in local storage.

PR #407, opened 2025-03-03
refactor: componentize `Conversation` elements

adamrturman:
## Summary
- refactors > 100 lines from `Conversation` into leaner, more singular-purpose components


downing034 (approved):
This seems fine. I did some spot checking and it looks decent as well.

I think in the future for similar re-organizing PRs, I'd break them into smaller chunks to make them easier to review, test, and isolate bugs if we have issues

PR #408, opened 2025-03-03
refactor: `Chatview` cleanup

adamrturman:
## Summary
- removed unnecessary prop drilling in favor of context
- made a `LandingContent` component for before a conversation starts

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,12 +7,10 @@ import { useIsMobile } from './shared/theme/useIsMobile';
 import { X as CloseX, PanelLeftClose } from 'lucide-react';
 import KavaAILogo from './core/assets/KavaAILogo';
 import ButtonIcon from './core/components/ButtonIcon';
-import { useMessageHistory } from './core/hooks/useMessageHistory';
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -92,14 +80,9 @@ export const App = () => {
 
           <div className={styles.content}>
             <ChatView
-              introText={modelConfig.introText}
-              cautionText={defaultCautionText}
-              messages={messages}
-              onSubmit={handleChatCompletion}
```
adamrturman, line 98: All of these values can be pulled directly from context or model config, so no need to drill them.

src/stories/ChatView.stories.tsx
```diff
@@ -21,27 +23,35 @@ const errorStore = new TextStreamStore();
 
 const withProviders =
   (
+    messages: ChatMessage[],
```
adamrturman, line 26: Only big adjustment had to come for storybook.

Each story can still set its own messages, just not via props, but via `messageHistoryStore.setMessages(...)`

samcode206 (approved):

PR #409, opened 2025-03-03
fix: verify both agent & view port to determine mobile

adamrturman:
## Summary

 When validating an evm tool call, we check that the user is not on mobile (and wouldn't be able to use metamask). However, if a user has the chrome dev tools open, they can get false positives based on the agent. Also, if a user resizes their desktop viewport to be narrower, we can also get false positives.
 
 So let's verify both. 

## Before

https://github.com/user-attachments/assets/9a01ca2e-75b1-45b8-bafe-25c1b3d322b6

## After

https://github.com/user-attachments/assets/d8dd3aab-d1ec-4f85-971e-d46913c48ba1

Screenshot 2025-03-03 at 12 53 18 PM

### DIFF
--- a/src/features/blockchain/services/evm/evmBalances.ts
+++ b/src/features/blockchain/services/evm/evmBalances.ts
@@ -23,10 +23,15 @@ export class EvmBalancesQuery implements ChainQuery<EvmBalanceQueryParams> {
   needsWallet = [WalletTypes.METAMASK];
 
   private isMobileDevice(): boolean {
-    // Use regex to detect mobile devices
-    return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
-      window.navigator.userAgent,
-    );
+    // use regex to detect mobile devices
+    const isMobileUserAgent =
+      /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
+        window.navigator.userAgent,
+      );
+
+    const isMobileViewport = window.innerWidth <= 768;
+
+    return isMobileUserAgent && isMobileViewport;
   }
 
   validate(params: EvmBalanceQueryParams, walletStore: WalletStore): boolean {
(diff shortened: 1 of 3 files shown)


src/features/blockchain/services/evm/evmBalances.ts
```diff
@@ -22,31 +23,9 @@ export class EvmBalancesQuery implements ChainQuery<EvmBalanceQueryParams> {
   walletMustMatchChainID = false;
   needsWallet = [WalletTypes.METAMASK];
 
-  private isMobileDevice(): boolean {
-    // Use regex to detect mobile devices
-    return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
-      window.navigator.userAgent,
-    );
-  }
-
```
adamrturman, line 31: Refactored to be shared

src/features/blockchain/services/evm/evmBalances.ts
```diff
@@ -22,31 +23,9 @@ export class EvmBalancesQuery implements ChainQuery<EvmBalanceQueryParams> {
...
-    if (Array.isArray(this.needsWallet)) {
-      if (!this.needsWallet.includes(walletStore.getSnapshot().walletType)) {
-        throw new Error('please connect to a compatible wallet');
-      }
-    }
-
-    if (!chainRegistry[this.chainType][params.chainName]) {
-      throw new Error(`unknown chain name ${params.chainName}`);
-    }
```
adamrturman, line 49: These three checks are also repeated and can be shared.

src/features/blockchain/utils/wallet.ts
```diff
@@ -0,0 +1,64 @@
...
+  const isMobileUserAgent =
+    /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
+      window.navigator.userAgent,
+    );
+
+  const isMobileViewport = window.innerWidth <= 768;
+
+  //  But also confirm the viewport (chrome devtools throws a false positive based on agent-alone)
+  return isMobileUserAgent && isMobileViewport;
+}
```
adamrturman, line 20: This is the main update here - don't just check the `userAgent`

downing034: I thikn we should move this viewport logic to the core space since it's not specific to the blockchain logic, just currently being used there.  I could also see a follow on ticket here to integrate this into the app itself.

PR #410, opened 2025-03-03
refactor: componentize sidebars

adamrturman:
## Summary
- App now returns components rather than some components JSX, so its easier to reason about purpose and structure.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,10 +4,10 @@ import { useAppContext } from './core/context/useAppContext';
 import styles from './App.module.css';
 import { ChatHistory } from './core/components/ChatHistory';
 import { useIsMobile } from './shared/theme/useIsMobile';
-import { X as CloseX, PanelLeftClose } from 'lucide-react';
 import KavaAILogo from './core/assets/KavaAILogo';
-import ButtonIcon from './core/components/ButtonIcon';
-import { SearchHistoryButton } from './core/assets/SearchHistoryButton';
+import { DesktopSideBar } from './core/components/DesktopSideBar';
+import { MobileSideBar } from './core/components/MobileSideBar';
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -25,9 +25,9 @@ export const App = () => {
     <>
       {isReady && (
         <div className={styles.app}>
-          <div
-            className={`${styles.backdrop} ${isMobileSideBarOpen ? styles.isOpen : ''}`}
-            onClick={() => setIsMobileSideBarOpen(false)}
+          <MobileBackdrop
+            styles={`${styles.backdrop} ${isMobileSideBarOpen ? styles.isOpen : ''}`}
+            onBackdropClick={() => setIsMobileSideBarOpen(false)}
```
adamrturman, line 30: Componentized this to show its purpose. At first I though this empty div could have been left over from a past refactor, but now its purpose is immediately clear.

src/App.tsx
```diff
@@ -25,52 +25,33 @@ export const App = () => {
...
           <div
             className={`${styles.sidebar} ${isMobileSideBarOpen ? styles.isOpen : ''} ${isDesktopSideBarHidden ? styles.isHidden : ''}`}
           >
             <div className={styles.sidebarHeader}>
               <KavaAILogo height={20} />
-              {isMobile && isMobileSideBarOpen && (
-                <div className={styles.buttonGroup}>
-                  <SearchHistoryButton
+              <div className={styles.buttonGroup}>
+                {isMobile && isMobileSideBarOpen && (
```
adamrturman, line 39: Both mobile and desktop versions are wrapped with `<div className={styles.buttonGroup} >` so let's move that outside the conditional and remove the repetition.

samcode206 (approved):

PR #416, opened 2025-03-04
feat: drag & drop UX 

adamrturman:
## Summary
- users can drag and drop supported files
- users are warned and blocked from uploading unsupported files

## Demo 

https://github.com/user-attachments/assets/664f5913-c4ab-4806-8ac7-ca8824a61abc

### DIFF
--- a/src/core/components/ButtonIcon.tsx
+++ b/src/core/components/ButtonIcon.tsx
@@ -12,6 +12,7 @@ interface TooltipProps {
   text: string;
   position?: TooltipPosition;
   delay?: number;
+  backgroundColor?: string;
 }
 
 export interface ButtonIconProps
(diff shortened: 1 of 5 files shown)


src/core/components/ButtonIcon.tsx
```diff
@@ -12,6 +12,7 @@ interface TooltipProps {
   text: string;
   position?: TooltipPosition;
   delay?: number;
+  backgroundColor?: string;
```
adamrturman, line 15: In order to render a tooltip that contrasted with the background, I needed to pass in a custom color

Screenshot 2025-03-04 at 4 01 33 PM

samcode206 (approved):

PR #418, opened 2025-03-05
refactor: use theme elements

adamrturman:
## Summary
- use theme spacing 
-  move remove image icon to top left
- refactor button disabled conditional to be more readable

## Demo 

Screenshot 2025-03-05 at 10 15 51 AM

### DIFF
--- a/src/core/components/ChatInput.module.css
+++ b/src/core/components/ChatInput.module.css
@@ -89,8 +89,8 @@
 }
 
 .uploadInputFieldContainer {
-  margin-left: 0.4rem;
-  margin-right: 0.4rem;
+  margin-left: var(--spacing-xs);
+  margin-right: var(--spacing-xs);
 }
 
 .uploadInputField {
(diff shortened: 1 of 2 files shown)


src/core/components/ChatInput.module.css
```diff
@@ -133,14 +133,14 @@
 
 .removeButton {
   position: absolute;
-  top: 0;
-  right: 0;
-  width: 2rem;
-  height: 2rem;
+  top: -0.1rem;
+  left: -0.2rem;
```
adamrturman, line 137: Adjusted so its tucked in the top left corner

src/core/components/ChatInput.tsx
```diff
@@ -85,10 +85,6 @@ const ChatInput = ({ setShouldAutoScroll }: ChatInputProps) => {
       return;
     }
 
-    if (inputValue === '' && imageIDs.length === 0) {
-      return;
-    }
-
```
adamrturman, line 91: This check is no longer necessary since the button that fires this function is disabled when this condition is true (verified via unit test).

https://github.com/Kava-Labs/oros/blob/6538e115370c9a75e6b78ead2090da6bd328a923/src/core/components/ChatInput.test.tsx#L45-L69

PR #426, opened 2025-03-06
test: image upload & chat rendering e2e

adamrturman:
## Summary
- verifies that a user can upload  single and multiple images and after submission, they are rendered in the conversation


e2e/chat.spec.ts
```diff
@@ -724,6 +728,14 @@ describe('chat', () => {
...
       expect(count).toEqual(1);
     }
+
+    await chat.submitMessage('Describe this image');
+
+    const uploadedImage = page.getByRole('img', {
+      name: 'User uploaded image 1',
+    });
+
+    await expect(uploadedImage).toBeVisible();
```
adamrturman, line 738: This is about as far as we can assert that won't flake in CI.

Locally, it works to wait for the reasoning model response to begin and then verify that the brain icon is rendered (essentially verifying that the model is reasoning about the image and didn't error), but that test times out everytime in CI/CD since the reasoning model is slow in that environment.

PR #437, opened 2025-03-06
feat: clear uploaded images on model switch and conversation switch

adamrturman:
## Summary
Image upload previews are reset when
- the user changes model
- the user changes conversation (even if they select an older conversation on reasoning model - state is fully reset)

## Demo 

https://github.com/user-attachments/assets/98336a44-d71f-4fbc-a981-4b0585615643

### DIFF
--- a/src/core/components/ChatInput.tsx
+++ b/src/core/components/ChatInput.tsx
@@ -70,6 +70,19 @@ const ChatInput = ({ setShouldAutoScroll }: ChatInputProps) => {
 
   const [imageIDs, setImageIDs] = useState<string[]>([]);
 
+  const prevModelIdRef = useRef(modelConfig.id);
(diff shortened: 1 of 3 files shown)


src/core/components/ChatInput.tsx
```diff
@@ -70,6 +70,19 @@ const ChatInput = ({ setShouldAutoScroll }: ChatInputProps) => {
...
+  useEffect(() => {
+    if (
+      prevModelIdRef.current !== modelConfig.id ||
+      prevConversationRef.current !== conversationID
+    ) {
+      setImageIDs([]);
+      prevModelIdRef.current = modelConfig.id;
+      prevConversationRef.current = conversationID;
+    }
+  }, [conversationID, modelConfig.id]);
```
adamrturman, line 85: If either of these values changes, reset the image previews.

PR #444, opened 2025-03-07
fix: Use a wrapper to center icons in line with each other

nddeluca:
This keeps the flex-end controls, but uses a wrapping container around only the button icons to keep them centered in a line with each other.

In addition, removes the margin in favor of gap.  When using flex to align items, gap is preferred to set spacing between elements.

## Demo

Before:
Screenshot 2025-03-06 at 4 58 58 PM

After:
Screenshot 2025-03-06 at 4 59 19 PM

### DIFF
--- a/src/core/components/ChatInput.module.css
+++ b/src/core/components/ChatInput.module.css
@@ -56,7 +56,13 @@
 .buttonContainer {
   flex-shrink: 0;
   display: flex;
+  align-items: flex-end;
+}
+
+.buttonWrapper {
+  display: flex;
   align-items: center;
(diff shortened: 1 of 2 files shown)


src/core/components/ChatInput.module.css
```diff
@@ -151,11 +157,6 @@
   background-color: var(--colors-bgSecondary);
 }
 
-
-.attachIcon {
-  margin-right: var(--spacing-xs);
```
nddeluca, line 156: Margin's are less flexible than setting flex gaps

samcode206 (approved):

PR #450, opened 2025-03-07
feat: support PDF uploads 

samcode206:
## Demo 

https://github.com/user-attachments/assets/614b3f7e-fdbd-4883-8f3d-fe5d2a779217


src/core/components/ChatInput.tsx
```diff
@@ -12,13 +12,21 @@ import { IdbImage } from './IdbImage';
 import ButtonIcon from './ButtonIcon';
 import { useTheme } from '../../shared/theme/useTheme';
 import { ChatCompletionContentPart } from 'openai/resources/index';
+import { PDFiumLibrary } from '@hyzyla/pdfium/browser/base64';
```
samcode206, line 15: note:  using the basr64 library is not recommended for prod.  https://pdfium.js.org/docs/init-library#browser

src/core/components/ChatInput.tsx
```diff
@@ -222,12 +230,31 @@ const ChatInput = ({ setShouldAutoScroll }: ChatInputProps) => {
...
+        console.log(e.target?.result);
         if (e.target && typeof e.target.result === 'string') {
           const imgID = await saveImage(e.target.result);
           setImageIDs((prevIDs) => [...prevIDs, imgID]);
+        } else if (e.target?.result instanceof ArrayBuffer) {
+          const buf = new Uint8Array(e.target.result);
+          const doc = await library.loadDocument(buf);
+
+          const page = doc.getPage(0);
+          const image = await page.render({ scale: 3, render: 'bitmap' });
```
samcode206, line 242: bitmap is the only possible option for the browser, we must take that and use canvas to convert to a base64image url.  sharp library may be used to directly convert to a png or possibly other formats but this is not possible to use in browsers.

adamrturman (approved):
Nice - if you rebase with main, the warning in `src/context/utils.ts` should be resolved

src/core/utils/pdf/pdf.ts
```diff
@@ -0,0 +1,69 @@
+import * as pdfjsLib from 'pdfjs-dist';
+
+pdfjsLib.GlobalWorkerOptions.workerSrc = `pdf.worker.min.mjs`;
```
nddeluca, line 3: Can we add a hash and version to the file name?

`pdf.worker.min-{version}-{short hash of file}.mjs`

PR #462, opened 2025-03-10
refactor: supported file types to model config

adamrturman:
## Summary
- implements the model config approach for supported file types set up in #430 
- removes deprecated/unused `isInIframe` helper

### DIFF
--- a/src/core/types/models.ts
+++ b/src/core/types/models.ts
@@ -69,4 +69,14 @@ export interface ContextMetrics {
   tokensRemaining: number;
 }
 
-type SupportedFileType = 'image/jpeg' | 'image/png' | 'image/webp';
+export const SUPPORTED_FILE_TYPES = [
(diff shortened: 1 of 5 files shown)


src/core/types/models.ts
```diff
@@ -69,4 +69,14 @@ export interface ContextMetrics {
...
+  'image/jpeg',
+  'image/png',
+  'image/webp',
+] as const;
+
+export type SupportedFileType = (typeof SUPPORTED_FILE_TYPES)[number];
+
+export function isSupportedFileType(type: string): type is SupportedFileType {
+  return SUPPORTED_FILE_TYPES.includes(type as SupportedFileType);
+}
```
adamrturman, line 82: Single touch point to add new file types like pdfs. This also abstracts away type conversion/casting out of the view layer.

nddeluca (approved):

PR #464, opened 2025-03-11
feat: qwen model behind feature flag

adamrturman:


src/features/reasoning/helpers.ts
```diff
@@ -40,7 +40,7 @@ export const estimateTokenUsage = (
   };
 };
 
-export async function calculateDeepseekTokenUsage(
+export async function calculateFinalChunkTokenUsage(
```
adamrturman, line 43: We can use this function to track qwen token usage by adding `stream_options: { include_usage: true }`.

That does impact how we parse the stream, since we will now have a final chunk with no content choices, but with a usage object. We need to account for that as a content chunk and not route it as a tool call as is the current schema.

src/core/utils/streamUtils.ts
```diff
@@ -7,12 +7,13 @@ import { ToolCallStreamStore } from '../stores/toolCallStreamStore';
  * @returns True if the chunk contains content, false otherwise.
  */
 export const isContentChunk = (result: ChatCompletionChunk): boolean => {
-  const delta = result.choices[0].delta;
-  // Sometimes content is an empty string, so we check if content is a string property.
-  if (delta && 'content' in delta && typeof delta.content === 'string') {
+  //  Treat usage-only chunks as content chunks
+  if (result.usage && (!result.choices || result.choices.length === 0)) {
```
adamrturman, line 11: Without this check, an error will be thrown because result.choices is an empty array in the final usage chunk from qwen.

src/core/context/utils.ts
```diff
@@ -226,17 +226,19 @@ export async function doChat(
...
       },
       {
         signal: controller.signal,
       },
     );
-
     let content = '';
     let thinkingEnd = -1;
+    let hasCloseTag = false;
+    let openTagFound = false;
```
adamrturman, line 238: Qwen can start without an opening think tag, but has a closing tag that we need to look for.

PR #465, opened 2025-03-11
fix: ensure that the initial system prompt is set

adamrturman:
## Summary
- In this bug fix #401 , we inadvertently removed setting the initial system prompt. Nothing triggered `ensureCorrectSystemPrompt` to run on initial load, only on model switch

## Before

https://github.com/user-attachments/assets/99789468-160e-435a-b5fc-dbf944a24414

## After

https://github.com/user-attachments/assets/88daf86d-53c2-4ef6-9497-0185244bcc98

### DIFF
--- a/src/core/context/AppContextProvider.tsx
+++ b/src/core/context/AppContextProvider.tsx
@@ -18,6 +18,7 @@ import { SupportedModels, ModelConfig } from '../types/models';
 import { ActiveConversation, ConversationHistory } from './types';
 import { getToken } from '../../core/utils/token/token';
 import OpenAI from 'openai';
+import { v4 as uuidv4 } from 'uuid';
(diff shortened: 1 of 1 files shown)


src/core/context/AppContextProvider.tsx
```diff
@@ -41,15 +41,49 @@ export const AppContextProvider = (props: {
...
+        const updatedMessages = [...messages];
+        updatedMessages[0] = {
+          role: 'system',
+          content: systemPrompt,
+        };
+        messageStore.setMessages(updatedMessages);
+      }
+    },
+    [],
+  );
```
adamrturman, line 70: Move up in the file to be referenced during state initialization.

samcode206 (approved):

PR #470, opened 2025-03-11
refactor: string literals for model names

adamrturman:


src/core/config/models.ts
```diff
@@ -19,12 +16,6 @@ export const isBlockchainModelName = (
   return Object.keys(MODEL_REGISTRY.blockchain).includes(name);
 };
 
-export const isReasoningModel = (
-  id: string,
-): id is SupportedReasoningModels => {
-  return Object.keys(MODEL_REGISTRY.reasoning).includes(id);
-};
-
```
adamrturman, line 27: Unused

PR #472, opened 2025-03-11
refactor: process upload file hook

adamrturman:


src/core/components/ChatInput.tsx
```diff
@@ -183,56 +184,14 @@ const ChatInput = ({ setShouldAutoScroll }: ChatInputProps) => {
...
-    [hasAvailableUploads, maximumFileBytes, resetUploadState],
-  );
+  const processUploadedFile = useProcessUploadedFile({
+    hasAvailableUploads,
+    maximumFileBytes,
+    setUploadingState,
+    saveImage,
+    setImageIDs,
+    resetUploadState,
+  });
```
adamrturman, line 194: Eventually, `saveImage` and `hasAvailableUploads` will also be isolated, so they can be imported and won't need to be arguments

PR #478, opened 2025-03-12
fix: resolve check

adamrturman:

### DIFF
--- a/src/core/utils/pdf/pdf.ts
+++ b/src/core/utils/pdf/pdf.ts
@@ -1,39 +1,10 @@
 import 'pdfjs-dist/build/pdf.worker.min.mjs';
 import * as pdfjsLib from 'pdfjs-dist';
-
-export async function pdfDocExtractText(
-  doc: ArrayBuffer,
-  nPages: number,
-): Promise<string[]> {
-  if (nPages <= 0) throw new Error(`nPages must be greater than 0`);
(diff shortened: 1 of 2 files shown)


src/core/utils/pdf/pdf.ts
```diff
@@ -1,39 +1,10 @@
 import 'pdfjs-dist/build/pdf.worker.min.mjs';
 import * as pdfjsLib from 'pdfjs-dist';
-
-export async function pdfDocExtractText(
```
samcode206, line 4: this is now being used

src/core/utils/pdf/pdf.ts
```diff
@@ -1,39 +1,10 @@
...
-
-  return pdfDocPerPageText;
-}
-
 export async function pdfDocToBase64ImageUrls(
   doc: ArrayBuffer,
   nPages: number,
 ): Promise<string[]> {
-  const loadingTask = pdfjsLib.getDocument(doc.slice());
+  const loadingTask = pdfjsLib.getDocument(doc);
```
samcode206, line 7: should use slice here, because the buffer's internal memory gets detached and cannot be accessed after it's passed to the pdf.js worker. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/detached

adamrturman: Cool - does it need args?

PR #482, opened 2025-03-14
feat: implement loading treatment for single and multiple file uploads

adamrturman:
## Summary
Eliminates the carousel being empty/collapsed before images are loaded

Screenshot 2025-03-14 at 10 45 21 AM

Screenshot 2025-03-14 at 10 45 33 AM


src/core/components/Conversation/Conversation.tsx
```diff
@@ -26,15 +26,11 @@ const ConversationComponent = ({ onRendered }: ConversationProps) => {
   );
 
   const { messages } = useMessageHistory();
-  // todo: re-implement this to make it work with the ImageCarousel
-  // const [loadedImages, setLoadedImages] = useState<Record<string, boolean>>({});
+  const [isLoaded, setIsLoaded] = useState(false);
```
adamrturman, line 29: We don't need to track loading states for individual files, just one state for when the DOM can be updated.

src/core/components/Conversation/ImageCarousel.tsx
```diff
@@ -26,13 +32,13 @@ export const ImageCarousel = ({ imageIDs }: ImageCarouselProps) => {
...
 
-  // Function for specific slide control
+  //  Function for specific slide control
   const currentSlide = (n: number) => {
     showSlides(n);
   };
 
   return (
-    <div>
+    <div className={isLoaded ? styles.visible : styles.hidden}>
```
adamrturman, line 41: The carousel is there the whole time, but once the first image is loaded and that state update is made, it will become visible.

PR #484, opened 2025-03-18
refactor: extract upload error

adamrturman:
## Summary
This sets up a new hook to manage setting and clearing the uploading error state. It also replaces two instances of such duplication (uploading too many files & uploading too big a single file). Follow on PR's can address other specific cases to allow for more focused testing.

## Demo

https://github.com/user-attachments/assets/79d195c0-3bf5-44c5-8dee-df578510a684

### DIFF
--- a/src/core/components/ChatInput.tsx
+++ b/src/core/components/ChatInput.tsx
@@ -344,6 +344,7 @@ const ChatInput = ({ setShouldAutoScroll }: ChatInputProps) => {
     modelSupportsUpload,
     resetUploadState,
     maximumFileBytes,
+    setUploadError,
   ]);
 
   const [, setHoverImageId] = useState<string | null>(null);
--- a/src/core/components/ChatInput.tsx
+++ b/src/core/components/ChatInput.tsx
@@ -325,16 +325,7 @@ const ChatInput = ({ setShouldAutoScroll }: ChatInputProps) => {
           const file = item.getAsFile();
           if (file) {
             if (file.size > maximumFileBytes) {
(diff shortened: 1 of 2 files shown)


src/core/components/ChatInput.tsx
```diff
@@ -249,15 +252,7 @@ const ChatInput = ({ setShouldAutoScroll }: ChatInputProps) => {
...
-        setUploadingState({
-          isActive: true,
-          isSupportedFile: false,
-          errorMessage: `Maximum ${maximumFileUploads} files allowed.`,
-        });
-
-        setTimeout(() => {
-          resetUploadState();
-        }, 2000);
+        setUploadError(`Maximum ${maximumFileUploads} files allowed.`);
```
adamrturman, line 255: Now the view just needs to specify the error message because the function returned from the hook handles the two state updates with the set timeout.

src/core/components/ChatInput.tsx
```diff
@@ -59,8 +60,10 @@ const ChatInput = ({ setShouldAutoScroll }: ChatInputProps) => {
     isSupportedFile: true,
     errorMessage: '',
   });
+  const { setUploadError } = useUploadingError(setUploadingState);
   const { isActive, isSupportedFile, errorMessage } = uploadingState;
 
+  //  todo - fully deprecate
   const resetUploadState = useCallback(() => {
```
adamrturman, line 67: When completely refactored, `resetUploadState` will only live inside `useUploadingError`

PR #490, opened 2025-03-26
feat: remove blockchain instruct from list of models

samcode206:


src/core/components/ModelSelector.tsx
```diff
@@ -121,6 +122,12 @@ export const ModelSelector = () => {
           onKeyDown={handleKeyDown}
         >
           {models.map((model, index) => {
+            if (model.name === BLOCKCHAIN_MODEL_NAME) {
+              // skip blockchain instruct model
+              // moving over to hard.ai
+              return null;
+            }
```
adamrturman, line 129: Couldn't we just remove the blockchainModels record from our supported models? Or are you thinking this PR does the UI removal only and then remove the config later?

PR #495, opened 2025-03-27
feat: drop tool call support, remove all related code

samcode206:
simplify model config

reorganize files to be less nested


e2e/chat.spec.ts
```diff
@@ -53,7 +52,11 @@ describe('chat', () => {
     expect(responseText).toMatch(/THIS IS A TEST/i);
   });
 
-  test.skip('check balances', async ({ page, context, metaMaskExtensionId }) => {
+  test.skip('check balances', async ({
```
adamrturman, line 55: I think we can just remove these, right?

src/core/types/models.ts
```diff
@@ -3,20 +3,24 @@ import {
...
-export type SupportedModels =
-  | SupportedBlockchainModels
-  | SupportedReasoningModel;
+export const supportedModels = [
+  'qwq-32b-bnb-4bit',
+  'deepseek-r1',
+  'gpt-4o',
+] as const;
+
+export type SupportedModels = (typeof supportedModels)[number];
```
adamrturman, line 15: Nice 😎

adamrturman (approved):

PR #497, opened 2025-03-28
feat: use o3-mini as the default reasoning model

samcode206:


src/core/utils/conversation/hasSufficientRemainingTokens.ts
```diff
@@ -9,14 +8,14 @@ import { estimateTokenUsage } from '../../config/models/helpers';
...
- * @param {SupportedModels} modelID - The identifier model being used.
+ * @param {string} modelID - The identifier model being used.
  * @param {string} input - The input text to be checked against available tokens
  * @param {number} availableTokens - The remaining number of available tokens in the conversation.
  *
  * @returns {boolean} - Returns true if the input is within the token limit, false otherwise.
  */
 export const hasSufficientRemainingTokens = (
-  modelID: SupportedModels,
+  modelID: string,
```
adamrturman, line 18: Did we lose the string literal type here?

PR #501, opened 2025-03-31
fix: fix e2e

adamrturman:


e2e/fixtures/utils.ts
```diff
@@ -16,11 +16,3 @@ export async function screenshotOnFailure(
...
   }
 }
-
-export function getWalletId(accountPool: number[], retry: number) {
-  if (retry) {
-    return accountPool[retry];
-  }
-
-  return accountPool[0];
-}
```
adamrturman, line 26: The rest of the metamask stuff will be removed in a follow on PR.

PR #507, opened 2025-04-02
refactor: use shared chat history helpers

adamrturman:


src/core/utils/conversation/helpers.test.ts
```diff
@@ -1,99 +1,10 @@
 import {
-  // calculateGptContextMetrics,
   formatContentSnippet,
-  formatConversationTitle,
-  getTimeGroup,
   groupAndFilterConversations,
   groupConversationsByTime,
 } from './helpers';
```
adamrturman, line 5: Note - we can't use the shared helpers here yet because of a currently different data shape here, but once we are moved from local storage to indexed db, we will be able to use them as well.

PR #508, opened 2025-04-02
refactor: lift up supports upload

adamrturman:

### DIFF
--- a/src/core/components/ChatView.tsx
+++ b/src/core/components/ChatView.tsx
@@ -11,12 +11,14 @@ export interface ChatViewProps {
   onMenu(): void;
   onPanelOpen(): void;
   isPanelOpen: boolean;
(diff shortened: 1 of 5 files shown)


src/core/components/ChatView.tsx
```diff
@@ -87,7 +89,10 @@ export const ChatView = ({
           >
             {!hasMessages && <LandingContent />}
 
-            <ChatInput setShouldAutoScroll={setShouldAutoScroll} />
+            <ChatInput
+              setShouldAutoScroll={setShouldAutoScroll}
+              supportsUpload={supportsUpload}
+            />
```
adamrturman, line 95: This component will be more easily reused if it can accept a prop to support uploads or not.

samcode206 (approved):

PR #523, opened 2025-04-04
fix: instantiate histories as null to fix flash in UI

adamrturman:
## Before
Since conversations was instantiated as an empty object, we got a flash of "no conversations" since we are checking its keys

https://github.com/user-attachments/assets/5dc7820f-0bc9-4a5b-b5fd-566bdf2a9b34

## After
Instantiate it as null, so we can use that as a base state, if no conversations exist it will eventually get set as an empty object 

Before loading

Screenshot 2025-04-04 at 2 07 09 PM

https://github.com/user-attachments/assets/fc57e393-91cb-463e-9cda-38a1dfbb14da


e2e/chat.spec.ts
```diff
@@ -41,382 +41,6 @@ describe('chat', () => {
     expect(responseText).toMatch(/THIS IS A TEST/i);
   });
 
-  test('chat history', async ({ page }) => {
```
adamrturman, line 44: This component is unit tested in the library

PR #525, opened 2025-04-07
refactor: containerize chat

adamrturman:
## Summary
Now that we've migrated off of context, we can continue the refactor of lifting state out of view components.

This first pass removes the chat view's subscription to the message history store and passes the message array instead of a store as props.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,10 @@
 import { useState } from 'react';
-import { ChatView } from './core/components/ChatView';
 import { useChat } from './core/context/useChat';
 import styles from './App.module.css';
 import KavaAILogo from './core/assets/KavaAILogo';
 import { SearchHistoryModal, SideBar, useIsMobileLayout } from 'lib-kava-ai';
 import { useSession } from './useSession';
+import { ChatViewContainer } from './core/components/ChatViewContainer';
 
 export const App = () => {
   const {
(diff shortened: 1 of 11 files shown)


src/core/components/ChatViewContainer.tsx
```diff
@@ -0,0 +1,65 @@
...
+  thinkingStore,
+  messageHistoryStore,
+  isRequesting,
+  errorStore,
+  messageStore,
+  startNewChat,
+  modelConfig,
+  conversationID,
+}: ChatViewContainerProps) => {
+  const { messages } = useMessageHistory(messageHistoryStore);
```
adamrturman, line 43: Subscribe to changes here so we don't re-render sidebar when new messages are added.

src/core/components/ModelSelector.tsx
```diff
@@ -2,37 +2,22 @@ import styles from './ModelSelector.module.css';
...
 
 export const ModelSelector = ({
   modelConfig,
-  messageHistoryStore,
+  isModelSelectorDisabled,
   handleModelChange,
 }: {
   handleModelChange: (modelName: SupportedModels) => void;
   modelConfig: ModelConfig;
-  messageHistoryStore: MessageHistoryStore;
```
adamrturman, line 14: No need to pass the whole store here - we just need to know when to disable it.

src/App.tsx
```diff
@@ -70,13 +66,9 @@ export const App = () => {
...
           />
           <div className={styles.content}>
-            <ChatView
+            <ChatViewContainer
               handleModelChange={handleModelChange}
-              thinkingStore={thinkingStore}
-              messageHistoryStore={messageHistoryStore}
-              isRequesting={isRequesting}
-              errorStore={errorStore}
-              messageStore={messageStore}
```
adamrturman, line 79: All of these are properties of the `activeChat`

src/core/components/ChatViewContainer.tsx
```diff
@@ -0,0 +1,58 @@
...
+    activeConversation.messageHistoryStore,
+  );
+
+  return (
+    <ChatView
+      handleModelChange={handleModelChange}
+      thinkingStore={activeConversation.thinkingStore}
+      isRequesting={activeConversation.isRequesting}
+      errorStore={activeConversation.errorStore}
+      messageStore={activeConversation.messageStore}
```
adamrturman, line 44: Follow on PR's to come to subscribe to these stores and pass down more primitive props

samcode206 (approved):

